PR #6, opened 2019-06-28
Enable default fault handler

steven-murray:
Enables the fault handler (same as recent PR for 21cmFAST).


steven-murray:
hey @BradGreig just pinging you to briefly review this

PR #13, opened 2019-09-04
fix pip installation files and add pipenv support

dprelogo:
Temporary fix of `pip` installation files.
For `package_data` location, I left it as `data/*`.

I've also added Pipfile for `pipenv`.
`pipenv install` creates new virtual environment and installs all packages.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,20 +4,13 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [dev-packages]
-
-[packages]
-click = "*"
-numpy = "*"
-cffi = "*"
-scipy = "*"
-astropy = "*"
 corner = "*"
-pytest = "*"
 pytest-cov = "*"
 pre-commit = "*"
 tox = "*"
-21cmfast = {git = "git://github.com/21cmFAST/21cmFAST"}
-PyYAML = "*"
+
+[packages]
+21cmfast = {editable = true,git = "git://github.com/21cmfast/21cmFAST"}
 21cmmc = {editable = true,path = "."}
 
 [requires]
--- a/Pipfile
+++ b/Pipfile
@@ -1,0 +1,24 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+
+[packages]
+click = "*"
+numpy = "*"
+cffi = "*"
+scipy = "*"
+astropy = "*"
+corner = "*"
+pytest = "*"
+pytest-cov = "*"
+pre-commit = "*"
+tox = "*"
+21cmfast = {git = "git://github.com/21cmFAST/21cmFAST"}
+PyYAML = "*"
+21cmmc = {editable = true,path = "."}
+
+[requires]
+python_version = "3.7"
--- a/setup.py
+++ b/setup.py
@@ -76,7 +76,7 @@ setup(
         "emcee<3",
         "powerbox>=0.5.7",
         "cached_property",
-        "21cmFAST @ git+git://github.com/21cmfast/21cmFAST#egg=21cmFAST",
+        "21cmFAST>=3.0.0dev",
     ],
     # entry_points={
     #     'console_scripts': [
--- a/setup.py
+++ b/setup.py
@@ -76,6 +76,7 @@ setup(
         "emcee<3",
         "powerbox>=0.5.7",
         "cached_property",
+        "21cmFAST @ git+git://github.com/21cmfast/21cmFAST#egg=21cmFAST",
     ],
     # entry_points={
     #     'console_scripts': [
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ setup(
     url="https://github.com/21cmFAST/21CMMC",
     packages=find_packages("src"),
     package_dir={"": "src"},
-    package_data={"py21cmmc": "data/*"},
+    package_data={"py21cmmc": ["data/*"]},
     py_modules=[splitext(basename(path))[0] for path in glob("src/*.py")],
     include_package_data=True,
     zip_safe=False,
@@ -76,7 +76,6 @@ setup(
         "emcee<3",
         "powerbox>=0.5.7",
         "cached_property",
-        "21cmFAST>=3.0.0dev",
     ],
     # entry_points={
     #     'console_scripts': [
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -76,7 +76,6 @@ def find_version(*file_paths):
         "emcee<3",
         "powerbox>=0.5.7",
         "cached_property",
-        "21cmFAST>=3.0.0dev",
```
steven-murray: I think this should be changed to `21cmFAST @ @ git+git://github.com/21cmFAST/21cmFAST`

Pipfile
```diff
@@ -0,0 +1,24 @@
...
+scipy = "*"
+astropy = "*"
+corner = "*"
+pytest = "*"
+pytest-cov = "*"
+pre-commit = "*"
+tox = "*"
+21cmfast = {git = "git://github.com/21cmFAST/21cmFAST"}
+PyYAML = "*"
+21cmmc = {editable = true,path = "."}
```
steven-murray, line 14: Since this is in the non-dev section, shouldn't 21CMMC be installed as not editable?

Pipfile
```diff
@@ -0,0 +1,24 @@
...
+[dev-packages]
+
+[packages]
+click = "*"
+numpy = "*"
+cffi = "*"
+scipy = "*"
+astropy = "*"
+corner = "*"
+pytest = "*"
```
steven-murray: I feel like these deps should be in `dev-packages`?
Also, should `cosmoHammer` be on this list somewhere?

steven-murray (commented):
Thanks so much @dprelogo ! This looks good -- just a couple of changes I requested.

I have never before used `pipenv`, though I have been meaning to use it and/or `poetry` for a while. I find that most people in science/data science use `conda`, which again has its own way of setting up an environment, which is why I have never gotten to it. It definitely doesn't hurt to support multiple methods of installation though!

setup.py
dprelogo: Ah yes, that's true... I've tried multiple formats, the link should just have this `egg` part also:
```suggestion
        "21cmFAST @ git+git://github.com/21cmfast/21cmFAST#egg=21cmFAST",
```

Pipfile
dprelogo: It is suggested on `pipenv` page to install all version-controlled packages as editable. I don't fully understand why, it has something to do with dependency checking. So `21cmfast` should also be installed as editable.

Pipfile
dprelogo: Ah yes, I was clumsy with what is dev and what is not...
For some reason I cannot attach new Pipfile, so here is the [link](https://www.dropbox.com/s/yqr0evy8rhcqo2s/pipfile?dl=0).

The only package needed for non-dev installation should be `21CMMC`, but I've also put `21cmfast`, as `pipenv` has some bug when reading links in `setup.py` and the package should also be editable (I think the two are connected).

After installation with this new Pipfile (so `pipenv install --dev`) here is an output of [`pipenv graph`](https://www.dropbox.com/s/yq565w6pw28srtj/pipenv_graph.txt?dl=0), so `cosmoHammer` and others are under `21CMMC`.

Pipfile
steven-murray: OK fair enough then :-)

steven-murray (approved):
Thanks @dprelogo !

Pipfile
steven-murray: OK I think I kind of understand: `pipenv` is using `setup.py` to get the deps of the package. But doesn't that just make `pipenv` redundant? Isn't the point of `pipenv` to be able to specify an exact environment (whereas requirements in `setup.py` specify unpinned requirements? 

I'm sorry I'm not up to speed with this! In the end, as long as it works for you and doesn't add extra complexity, I'm OK with it!

setup.py
steven-murray: Ah OK, that looks right!

steven-murray:
It looks like travis is failing and saying that direct URLs are not allowed as dependencies. They are definitely allowed in the `install_requires` in `setup.py` -- I'm not sure which part they're not allowed in.

dprelogo:
I've read somewhere that links in `setup.py` are not allowed for a release on PyPI, I'm not sure if that's why it fails.

Just as a note, I've also tried the same thing with [`dependency_links`](https://python-packaging.readthedocs.io/en/latest/dependencies.html), but it seems that's deprecated.

Pipfile
dprelogo: One can think of `pipenv` as `pip`, with addition that it creates virtual environment, manages all the venvs and strictly checks dependency compatibilities. 
I actually haven't realised at first how nicely have you packed everything, so except for easily creating dev environment, it is redundant.

steven-murray: I'm happy for it still to be there just for the dev environment. Doesn't it do stuff like automatically update your requirements when you update packages? That seems useful enough.

steven-murray:
Yeah `dependency_links` are not a thing anymore. I don't _think_ this is a problem with it being on PyPI, because it shouldn't be installing it from there... I'll have a bit of a poke around.

steven-murray:
Can you try changing `requirements_dev.txt` so that it has `'21cmFAST @ git+git://github.com/21cmFAST/21cmFAST',`?

Pipfile
dprelogo: Yes, it does that! OK, great :)

dprelogo:
It is definitely the one in `setup.py`, it seems...

steven-murray:
It's very strange, because it works perfectly fine in py3.7...

steven-murray:
Having thought about this, I think we want to take out the direct URL here, and just leave it as `21cmFAST>=3.0.0dev`. This doesn't exist yet, but it should exist in the next few weeks. Installing from `requirements_dev.txt` works, and that's what we should all be doing at the moment.

Long-term, we want people to be installing the actual PyPI release, not the github release. Even for developing, we don't want to merge code that _depends_ on the latest github release -- we need to be relying on the PyPI releases that everyone else will be getting. 

So I say change it back to not a github link. It will mean that you need to install (as a developer) from the `requirements_dev.txt` for now, until we make our first PyPI release.

dprelogo:
OK yes, I would also say that is a proper way...

steven-murray (approved):
Thanks @dprelogo this looks great!

PR #17, opened 2019-09-24
Add capability to use OR in checking required cores

### LINKED ISSUE #10: [Feature Req.] Clarification of CoreModule verification in LikelihoodModules
In each of the different likelihood modules already written, the verification of the provided linked core is different. 
Some of them are done in the setup, in others, the check is done directly in the compute_likelihood, or even in the reduce data. Also, because of this variety of methods, I do not understand how the required_core works. In most of them, I think it is not used.

It would be better to have one fix method to check for the needed cores and stick to it.

### LINKED ISSUE #20: [BUG] Tutorial error
**Describe the bug:**

Minor bug in mcmc_intro tutorial.  

`from py21cmmc.mcmc import analyse` should be `from py21cmmc import analyse` (I think)

**To Reproduce:**
Run the first import cell.

steven-murray:
Fixes #10 
Fixes #20 

Cleans up the checking of `required_cores` somewhat, adding the ability to have the equivalent of "or", so you can specify that a `Likelihood` requires `CoreThis or CoreThat`, via sub-tuples.


ghost (approved):
I can't see any obvious problems.

PR #19, opened 2019-10-04
adding LikelihoodEDGES

qyx268:
include features that allow one to compare against EDGES detection


steven-murray:
HI @qyx268, thanks for making this. Overall, you've constructed the Likelihood basically as it should be, though there's a few things I will comment on. Firstly though, can I ask what the purpose of this is? Why don't you just compare directly to the "best-fit" model of EDGES, or something like that? Why is your single number the timing? Is that because we may only trust that part?

It's not clear that the likelihood actually makes sense -- the timing is probably not a Gaussian variable. However, I guess it practically makes some sense. I'll review the code specifically soon.

qyx268:
Hi @steven-murray , thanks for following up on this. The reason that we only use the timing (and FWHM) is that the current detection cannot be explained by standard 21cm models. Our model can now be constrained by its timing, which might suggest efficient star formation in the first galaxies. We will include more physics in the future to explain other features including the FWHM and amplitude of EDGES detection. But still, we do not have a model which can explain its flattening, which is the main reason that we decided to not use the whole spectra of EDGES.

src/py21cmmc/__init__.py
```diff
@@ -15,6 +15,8 @@
     LikelihoodLuminosityFunction,
     LikelihoodPlanck,
     Likelihood1DPowerLightcone,
+	LikelihoodEDGEStimingOnly,
```
steven-murray: I think this is using a <TAB> but it should be four spaces. Ensure your editor is set-up to insert spaces for tabs!

src/py21cmmc/likelihood.py
```diff
@@ -956,3 +956,101 @@ def define_noise(self, ctx, model):
             return [{"sigma": s(m["Muv"])} for s, m in zip(sig, model)]
         else:
             return [{"sigma": s} for s in sig]
+
+
+class LikelihoodEDGES(LikelihoodBaseFile):
+    """
+    A likelihood based on chi^2 comparison to Global Signal of EDGES timing and FWHM
```
steven-murray: It would be good to include a more thorough description of the actual likelihood in the docstring here, and also what the parameters are that can be modified.

src/py21cmmc/likelihood.py
```diff
@@ -956,3 +956,101 @@ def define_noise(self, ctx, model):
...
+    FREQ_ERR_EDGES = 1.
+    FWHM_EDGES = 19.
+    FWHM_ERR_UPP_EDGES = 4.
+    FWHM_ERR_LOW_EDGES = 2.
+
+    required_cores = [core.CoreLightConeModule]
+
+    def reduce_data(self, ctx):
+        frequencies=1420.0 / (np.array(ctx.get("lightcone").node_redshifts) + 1),
+        global_signal=ctx.get("lightcone").global_brightness_temp,
```
steven-murray: trailing comma will ruin things

src/py21cmmc/likelihood.py
```diff
@@ -956,3 +956,101 @@ def define_noise(self, ctx, model):
...
+    FREQ_EDGES = 78.
+    FREQ_ERR_EDGES = 1.
+    FWHM_EDGES = 19.
+    FWHM_ERR_UPP_EDGES = 4.
+    FWHM_ERR_LOW_EDGES = 2.
+
+    required_cores = [core.CoreLightConeModule]
+
+    def reduce_data(self, ctx):
+        frequencies=1420.0 / (np.array(ctx.get("lightcone").node_redshifts) + 1),
```
steven-murray: trailing comma will ruin things

src/py21cmmc/likelihood.py
```diff
@@ -956,3 +956,101 @@ def define_noise(self, ctx, model):
...
+        global_signal=ctx.get("lightcone").global_brightness_temp,
+        f = InterpolatedUnivariateSpline(frequencies, global_signal,k=4)
```
steven-murray: please use a more descriptive variable name!

src/py21cmmc/likelihood.py
```diff
@@ -956,3 +956,101 @@ def define_noise(self, ctx, model):
...
+        cr_pts = f.derivative().roots()
```
steven-murray: also a more descriptive name!

src/py21cmmc/likelihood.py
```diff
@@ -956,3 +956,101 @@ def define_noise(self, ctx, model):
...
+        cr_vals = f(cr_pts)
```
steven-murray: more descriptive name, and/or a comment explaining what these are

src/py21cmmc/likelihood.py
```diff
@@ -956,3 +956,101 @@ def define_noise(self, ctx, model):
...
+        if (len(cr_vals) == 0):
+            return dict(Freq_Tbmin = None, FWHM = None)
+        else:
```
steven-murray, line 1011: this "else" is not required, and you can un-indent the rest

src/py21cmmc/likelihood.py
```diff
@@ -956,3 +956,101 @@ def define_noise(self, ctx, model):
...
+            Freq_Tbmin = cr_pts[np.argmin(cr_vals)]
```
steven-murray: use only lower-case letters for variable names

src/py21cmmc/likelihood.py
```diff
@@ -956,3 +956,101 @@ def define_noise(self, ctx, model):
...
+            Freqs_HM= InterpolatedUnivariateSpline(frequencies,global_signal-f(Freq_Tbmin) * 0.5, k=3).roots()
+            if len(Freqs_HM) == 2:
```
steven-murray: please add a comment to each of these cases explaining what it means

src/py21cmmc/likelihood.py
```diff
@@ -956,3 +956,101 @@ def define_noise(self, ctx, model):
...
+            else:
+                return dict(Freq_Tbmin = Freq_Tbmin, FWHM = Freq_r - Freq_l)
+
+    def computeLikelihood(self, model):
+        """
+        Compute the likelihood, given the lightcone output from 21cmFAST.
+        """
+
+        if model["Freq_Tbmin"] is None or model["FWHM"] is None:
+            return -10000000000.
```
steven-murray: return `-np.inf`

src/py21cmmc/likelihood.py
```diff
@@ -956,3 +956,101 @@ def define_noise(self, ctx, model):
...
+        else:
+            return -0.5 * np.square((model["Freq_Tbmin"] - self.FREQ_EDGES ) / self.FREQ_ERR_EDGES) +\
```
steven-murray: I think it's nicer to have two lines, something like:
```
timing_lnl = -0.5 * np.square((model["Freq_Tbmin"] - self.FREQ_EDGES ) / self.FREQ_ERR_EDGES)
width_lnl = -0.5 * np.square(model["FWHM"] - self.FWHM_EDGES) / (self.FWHM_ERR_UPP_EDGES * self.FWHM_ERR_LOW_EDGES + (self.FWHM_ERR_UPP_EDGES - self.FWHM_ERR_LOW_EDGES) * ( model["FWHM"] - self.FWHM_EDGES))
return timing_lnl + width_lnl
```
However, also make sure that the line length is not too long

src/py21cmmc/likelihood.py
```diff
@@ -956,3 +956,101 @@ def define_noise(self, ctx, model):
...
+                   -0.5 * np.square(model["FWHM"] - self.FWHM_EDGES) / (self.FWHM_ERR_UPP_EDGES * self.FWHM_ERR_LOW_EDGES + (self.FWHM_ERR_UPP_EDGES - self.FWHM_ERR_LOW_EDGES) * ( model["FWHM"] - self.FWHM_EDGES))
+
+class LikelihoodEDGEStimingOnly(LikelihoodBaseFile):
```
steven-murray: I think you can probably dispense with this class, and just use the previous one, with an option `use_width=False`.

tests/test_mcmc.py
```diff
@@ -381,6 +381,29 @@ def test_global_signal(lc_core, lc_core_ctx):
     assert "frequencies" in model
 
 
+def test_edges_timing_only(lc_core, lc_core_ctx):
```
steven-murray, line 378: this test will need to be modified if you use only the single class with an option, as I mentioned above.

steven-murray (commented):
Thanks @qyx268! I understand the idea here, and it is a good idea -- I've made a few comments on the code itself. After fixing those, it should be good to go!

qyx268:
Can someone review this change please?

steven-murray (approved):
This looks great! Thanks @qyx268

steven-murray:
I'll let you merge and delete the branch :-)

PR #23, opened 2019-10-24
Flake8

steven-murray:
This PR adds a number of new formatting checks and auto-corrections, as well as actually tidying the code up using these corrections. 

All of these checks are available via `pre-commit`, and are also run via CI/tox.


requirements_dev.txt
```diff
@@ -1,12 +1,13 @@
-click
-numpy
-pyyaml
-cffi
-scipy
+21cmFAST>=3.0.0dev
```
JulianBMunoz, line 1: The syntax of this is slightly different than before, but I'm going to assume it's right.
I see that you added and removed the 21cmFAST @ git line.

setup.py
```diff
@@ -68,7 +68,6 @@ def find_version(*file_paths):
     keywords=["Epoch of Reionization", "Cosmology"],
     install_requires=[
         "click",
-        # 'tqdm',
```
JulianBMunoz, line 71: We don't use progress bars? tqdm is a pretty good package for that. At any rate if it was commented out I guess we don't need it.

tests/test_mcmc.py
```diff
@@ -76,16 +81,14 @@ def test_core_coeval_setup(core, likelihood_coeval):
     assert not np.all(ctx.get("brightness_temp") == 0)
 
 
-def test_mcmc(core, likelihood_coeval, tmpdirec):
+def test_mcmc(core, likelihood_coeval, default_params, tmpdirec):
```
JulianBMunoz, line 84: much better syntax!

JulianBMunoz (commented):
Hey! All the style changes look good to me. The only thing I'm a little confused about is the requirements file, since the syntax called for github but now just requires 21cmFASTv3dev or later, right?

Bear in mind that I haven't used the python part of 21cmmc at all, so other than the syntax I can't say much. If that's OK with you i'm happy to approve.

requirements_dev.txt
steven-murray: Yes, this is something we should be doing -- that is, only relying on the most obvious _publicly available_ release of 21cmFAST. Installing from GitHub means we'll always get all the latest bits and pieces that haven't been formally released yet.

setup.py
steven-murray: No, we don't at the moment! We should, but it's not there yet, so no reason to include it.

steven-murray:
Yeah, there's not much in this PR except for syntax :-)

Feel free to approve!

JulianBMunoz (approved):
All changes since last time look good to me!

JulianBMunoz (approved):

JulianBMunoz (approved):

PR #26, opened 2019-11-22
Memory leak

### LINKED ISSUE #4: [BUG] Memory Leak
This is a copy/restatement of part of https://github.com/BradGreig/Hybrid21CM/issues/30.

**Describe the bug:**

@caw11 stated:

> Just to add to this, in case our prior ranges are equally flawed.

> I have some students running the higher parameterisation (i.e. that of Park+ 2018 but without spin temp) and are experiencing a similar problem. Our params dictionary looks like this:

> ```FitParams = dict(F_STAR10 = [-1.3, -3.0, 0.0, 0.1],
> ALPHA_STAR = [0.5, -0.5, 1.0, 0.05],
> F_ESC10 = [-1.0, -3.0, 0.0, 0.1],
> ALPHA_ESC = [-0.5, -1.0, 0.5, 0.05],
> M_TURN = [8.7, 8.0, 10.0, 0.1],
> t_STAR = [0.5, 0.0, 1.0, 0.05]).```

> With 8 threads the code consistently crashes after ~40 iteration with it sometimes returning the same exception as described above. However, they have observed that if they run in iterations batches of 30 continuing sampling, then it stalls much less regularly. I wonder if there might be a memory leak issue of some sort on top of the discontinuity issues in some regions of parameter space? Apologies if such a memory leak issue has already been raised elsewhere.

@BellaNasirudin added:
> I have been (still am) experiencing memory leak issue as well. I am using this version commit 450d096

> In my case, it seems to come from the storage module in 21cmmc since when I opted out of doing that, the code ran through perfectly fine and only used 100 GB. Of course, the downside is that I cannot analyse the walkers if there is something wrong with the MCMC and will have to run everything again.

> With the storage module included, I am using at least 915 GB so it's very computationally expensive.

**Expected behavior:**

That memory would not grow so large as to crash the system.

steven-murray:
Adds some tools to check memory leaks in 21cmFAST as they appear in 21cmMC.
Also fixes access to the new `Coeval` object in 21cmFAST.
Fixes #4 .


src/py21cmmc/core.py
```diff
@@ -273,11 +273,11 @@ def __init__(
             are required to specify the ionization field. Beyond this, the ionization field is specified directly from
             the perturbed density field.
         ctx_variables : list of str, optional
-            A list of strings, any number of the following: "brightness_temp", "init", "perturb", "xHI". These each
-            correspond to an OutputStruct which will be stored in the context on every iteration. Omitting as many as
+            A list of strings. The strings must correspond to any (pickleable) member of
+            :class:`py21cmfast.Coeval`. These will be stored in the context on every iteration. Omitting as many as
```
catherinewatkinson, line 277: Add the strings that users can choose from. Make life easy for us :)

src/py21cmmc/likelihood.py
```diff
@@ -930,42 +930,117 @@ class LikelihoodLuminosityFunction(LikelihoodBaseFile):
...
         lnl = 0
         for i, z in enumerate(self.redshifts):
             model_spline = InterpolatedUnivariateSpline(
                 model["Muv"][i][::-1], model["lfunc"][i][::-1]
             )
-
             lnl += -0.5 * np.sum(
-                (self.data["lfunc"][i] - model_spline(self.data["Muv"][i])) ** 2
-                / self.noise[i]["sigma"] ** 2
+                (self.data["lfunc"][i] - 10 ** model_spline(self.data["Muv"][i])) ** 2
```
catherinewatkinson: I'm not familiar with the subtleties of the LF likelihood, so another pair of eyes should check that this is correct.

catherinewatkinson (approved):
I have added a couple of comments to the change summaries that need attention. I will approve before this though as I know it is urgent.

src/py21cmmc/core.py
steven-murray: So the problem with adding the strings here is that the list might get out of date. What I might do is list a few here as typical suggestions, and refer the user to a function they can run to get the full current list of possibilities.

src/py21cmmc/likelihood.py
steven-murray: Don't worry about this -- it was merged to this branch by accident by @binnietom. It will be reviewed in its own PR (#32 )

steven-murray:
Thanks @catherinewatkinson! I've emended the branch so that @binnietom 's merge commit is no longer in it (so no need to check the LF stuff here). I've also listed a few common options for the `ctx_variables`, without suggesting it's an exhaustive list. Actually obtaining an exhaustive list is actually kind of annoying right now for the user. I'm going to make an issue right now on 21cmFAST to add an easy function for doing so. In the meantime, the user can create a Coeval cube and check its atrributes.

catherinewatkinson:
It doesn't have to be exhaustive... just add the main ones and tell them how they can create a coeval and check the rest themselves.

catherinewatkinson:
re. your point about this info getting out of date. Surely any change to how the user interfaces at a high level should only change when you do a new version (they certainly shouldn't if you want people to keep using the code). At which point, you also should be updating the documentation and stopping it getting out of date.

steven-murray:
That's right, but this wouldn't change the user interface -- adding an option which is off by default doesn't break anyone's code. You don't want to have to search for ten places you documented something everytime you change it, if you can help it :-)

Issue #25, opened 2019-11-12
MultiNest sampler


qyx268:
**Is your feature request related to a problem? Please describe.**
to reach convergence faster

**Describe the solution you'd like**

to include an additional sampler to emcee -- multinest

**Describe alternatives you've considered**

**Additional context**

PR #27, opened 2019-11-27
adding Multinest sampler

qyx268:
This is related to Issue #25 where we've added the multinest sampler

@steven-murray 
I don't see a decent way to output data (e.g. global_xH or PS) using multinest. Unlike cosmoHammer, I think it's not very straightforward to pass a storageUtil class. Do you have any suggestion? I'm thinking of outputting inside the likelihood function when calling multinest?


qyx268:
image

Test with LikelihoodNeutralFraction, LikelihoodGreig, LikelihoodPlanck. Blue and Red correspond to Multinest and Emcee samplers, respectively. The posteriors look similar between the two. 

MultiNest has 582 total likelihood evaluations before reaching convergence with ln(ev)=  -2.09      +/-  0.12. The Emcee result shown in the posterior has 9000 evaluations. But from the chain, it looks like it has converged after 6000 evaluations.

src/py21cmmc/likelihood.py
```diff
@@ -965,3 +965,122 @@ def define_noise(self, ctx, model):
...
+
+    Parameters
+    ----------
+
+    use_width : bool
+        whether to use the FWHM in the likelihood, by default it's False
+
+    """
+
+    FREQ_EDGES = 78.
```
steven-murray: To keep with Python style guidelines, these should be all lower-case

src/py21cmmc/likelihood.py
```diff
@@ -965,3 +965,122 @@ def define_noise(self, ctx, model):
...
+    def reduce_data(self, ctx):
+        frequencies=1420.0 / (np.array(ctx.get("lightcone").node_redshifts) + 1)
+        global_signal=ctx.get("lightcone").global_brightness_temp
+        global_signal_interp = InterpolatedUnivariateSpline(frequencies, global_signal,k=4)
+        cr_pts = global_signal_interp.derivative().roots()
+        cr_vals = global_signal_interp(cr_pts)
+        if (len(cr_vals) == 0):
+            return dict(Freq_Tbmin = None, FWHM = None) # there is no solution -- global signal never reaches a minimum
+        else:
+            Freq_Tbmin = cr_pts[np.argmin(cr_vals)]
```
steven-murray: change this variable to ``freq_tb_min``.

src/py21cmmc/likelihood.py
```diff
@@ -965,3 +965,122 @@ def define_noise(self, ctx, model):
...
+            if not self.use_width:
+                return dict(Freq_Tbmin = Freq_Tbmin,  FWHM = None)
+            # calculating the frequencies when global signal = the FWHM
+            Freqs_HM= InterpolatedUnivariateSpline(frequencies,global_signal-global_signal_interp(Freq_Tbmin) * 0.5, k=3).roots()
```
steven-murray: again, change all variables to be lower case, except for class names

src/py21cmmc/mcmc.py
```diff
@@ -180,8 +180,144 @@ def run_mcmc(
...
+    log_level_21CMMC=None,
+    n_live_points=100,
+    importance_nested_sampling=True,
+    sampling_efficiency=0.8,
+    evidence_tolerance=0.5,
+    max_iter=50,
+    multimodal=True,
+    write_output=True,
+    resume=True,
+    **multinest_options,
```
steven-murray: The "Essential" multinest options should be put in here explicitly, and the rest can be caught by the ``**multinest_options`` here.

src/py21cmmc/mcmc.py
```diff
@@ -180,8 +180,144 @@ def run_mcmc(
...
+        The logging level of the cosmoHammer log file.
+        for details. This setting affects only the file output by the MCMC run.
+    log_level_mcmc_stream : (int or str, optional)
+        The logging level of the stdout/stderr stream from cosmoHammer. This has the same output as the cosmoHammer
+        log file, but is printed to screen. See `log_level_mcmc` for input specifications.
+    log_level_21CMMC : (int or str, optional)
+        The logging level of the 21cmFAST Python code (specifically the "21CMMC" logging object). By default, this
+        logger has only a stdout handler. See https://docs.python.org/3/library/logging.html#logging-levels
+
+    Essential MultiNest Parameters
```
steven-murray: The parameters here should be added to the ``Parameters`` section (as noted in previous comment)

src/py21cmmc/mcmc.py
```diff
@@ -180,8 +180,144 @@ def run_mcmc(
...
+
+    Other Parameters
+    ----------------
+    All other parameters are passed directly to multinest
+
+    Returns
+    -------
+    sampler : multinest sampler
+    """
+    import pymultinest
```
steven-murray: do this as 

```python
try:
    import pymultinest
except ImportError:
    raise ImportError("You need to install pymultinest to use this function!")
```

steven-murray: Also, add an ``extras_require`` in ``setup.py`` something like:

```python
setup(
    ...
    extras_require = {
        "samplers": ['multinest'],
    }
)
```

src/py21cmmc/mcmc.py
```diff
@@ -180,8 +180,144 @@ def run_mcmc(
...
+
```
steven-murray: From here to line 299 is very similar to `run_mcmc`, and we should probably refactor so we don't have redundant code.

steven-murray (commented):
This is a great addition, thanks @qyx268 !

Your question about writing out "blobs" is basically the reason I haven't tried to do this myself in the past -- it seems to me like a tricky question. In fact, I'm not even really happy with the way that `emcee` writes out blobs right now (I've had to copy in code from `emcee v3` but modify it to get it to work). It's method of course only works for `emcee`, since it happens in the sampler itself, so it's not translatable to other samplers. 

We could imagine using a similar HDF5-backed system, but including it explicitly in the likelihood, so that other samplers work with it, as you suggest. However, I think that will take a little bit of thought, and we will want to get it basically right first time (so that we don't have to change the API too much). 

We have two options: (1) we can think about how to get the blobs right for different samplers right here in this PR, which will delay the acceptance of this PR, but means the blob saving will hopefully get into 21CMMC v1; (2) we can accept this PR as-is, and treat the proper writing out of blobs as an extra feature, for which we should make an issue, and postpone it to 21CMMC v1.1 or even v2. 

I'll invite @BradGreig and @caw11 to comment on which option they think is best.

src/py21cmmc/mcmc.py
qyx268: I don't think I understand this. These parameters are taken directly from MultiNest, which should be specified when calling run_multinest. Therefore I consider them as default Parameters, same as core_modules, likelihood_modules, etc. Putting them in a second section (Essential MultiNest Parameters) is just showing that they are from MultiNest.

Are you suggesting removing the title (Essential MultiNest Parameters) and put these parameters in the "Parameter" section, or putting these parameters in **essential_multinest?

src/py21cmmc/likelihood.py
qyx268: These are for #19.

src/py21cmmc/mcmc.py
qyx268: Are you suggesting simply renaming them or putting run_multinest as an option for run_mcmc?

qyx268:
I would leave the decision to you whether to include this PR in the following release or to postpone it when outputting functions are included. I'm okay with both. But as you can see, this branch (for PR #27 ) was created from the mini-halos branch (for PR #19 ) where EDGES likelihood was included. I would suggest accepting #19 first and include EDGES likelihood in the v1. 

I also feel that the best way would be outputting directly in the likelihood. This is what I had in mind but relevant development would take much longer. 

I'm pushing this implementation because we might have a 2-month lifetime computing resource in a few weeks where I'd like to test the scalability of this sampler. For this, I might simply output tiny files for blobs as a temporary solution.

qyx268:
Here is an updated comparison with 1000 lives points when running MultiNest (compared to 100 last time). It has 5892 evaluations before convergence. Again Multinest and EMCEE are indicated in blue and red, respectively. The values shown on the top of the 1D PDF are median+/-[14,86]percentiles (maximum likelihood position)

I've also updated the EMCEE results with the last 5892 evaluations. Clearly we see similar posterior distribution with MultiNest being broader than EMCEE result. This is something worth looking into details.

image

qyx268:
@BradGreig  I remember before when I showed you the comparison between Emcee and MultiNest using 5 parameters, it's also true that the posteriors look similar with MultiNest being a bit broader. You had the concern about sampling -- the number of points used to calculate the posterior is not large enough (I was using 7663 for the 5-parameter runs). I think, in this 3-parameter result where 1000 lives points were adopted and ~6000 points are used in the posterior, the factor that the comparison has the same behaviour indicates that Emcee might be over converged?

src/py21cmmc/likelihood.py
```diff
@@ -965,3 +965,139 @@ def define_noise(self, ctx, model):
...
+    fwhm_err_low_edges = 2.0
+
+    required_cores = [core.CoreLightConeModule]
+
+    def __init__(self, use_width=False, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.use_width = use_width
+
+    def reduce_data(self, ctx):
+        frequencies = 1420.0 / (np.array(ctx.get("lightcone").node_redshifts) + 1)
```
BradGreig: This should be 1420.4

I know it won't change any results, but it should at least be accurate. Probably should explicitly be 1.0 not 1, not sure how much is matters in Python for adding different variable types (or if it is intelligent enough to know that its a double)

src/py21cmmc/likelihood.py
```diff
@@ -965,3 +965,139 @@ def define_noise(self, ctx, model):
...
+        if model["freq_tb_min"] is None:
+            return -np.inf
+
+        if self.use_width:
+            if model["fwhm"] is None:
+                return -np.inf
+            else:
+                return -0.5 * np.square(
+                    (model["freq_tb_min"] - self.freq_edges) / self.freq_err_edges
+                ) + -0.5 * np.square(model["fwhm"] - self.fwhm_edges) / (
```
BradGreig: You need to be careful when using this approximation for asymmetric errors. You need to put in checks that the denominator doesn't go negative

BradGreig:
@qyx268 It may not be that its over converged, but perhaps dependent on what are the assumed initial conditions for EMCEE. I believe the default is to sample the initial conditions around the locations of the fiducial parameters. If you were to use completely random initial conditions instead, I wonder if the results would change slightly. 

In any case for nested sampling you must sample additional models in order to better fill out the posteriors for drawing contours. To me this comparison is pretty good.

src/py21cmmc/likelihood.py
qyx268: can I confirm with you that it is 1419.4697 or 1420.4?

BradGreig: It is 1420.405751...

qyx268: of course... now I'm wondeing why I've been using 1419 since the beginning ...

steven-murray:
OK, let's wait until #19 is merged, and we can merge this one in v3.1 or so. Of course, you're more than welcome to use it for applications before then, with any hack you have for writing out blobs!

If we want to support different samplers into the future (emcee, multinest, polychord,...), it may be good to have a generic way of saving parameters/blobs from within the likelihood itself, in case the particular sampler doesn't expose their parameter sampling method. The thing is, the idea _should_ be to subclass the `CosmoHammerSampler` class for each kind of sampler. This can still be done (even for multinest), but because multinest is written in fortran(?), we don't have access to the actual parameters/outputs on each iteration -- it just runs away and does its thing. Polychord does the same thing.

qyx268:
great, can someone review #19 please, is there anything else I need to do?

steven-murray:
@qyx268 where's this PR at? If we update the branch with the base branch, could we merge it in?

qyx268:
I have the impression that we've postponed it until outputting within likelihood is included. However, this PR does allow users to have access to parameters in the posterior. We can merge it now and users can hack the likelihoods for output if needed. 

I just merged the master into this branch. But it seems that, on Travis, `external_table_path`
is not correctly set, leading crashes in most tests

src/py21cmmc/mcmc.py
steven-murray: You can put them in a "Other Parameters" section -- remember that these docstrings are parsed by a program to convert to html documentation. The parser knows about the Parameters section and an "Other Parameters" section, and can format them appropriately.

src/py21cmmc/mcmc.py
steven-murray: I think I would favour the latter, actually. Let's just add an option for `run_mcmc` like `sampler`, with `emcee` as the default, but one could specify `"multinest"` if they want.

steven-murray:
@qyx268 OK thanks for the update! I say we merge it without the ability to do the full outputting within the likelihood, and create an issue for that right now, so we don't forget to add the feature later. 

You should also make a (small) tutorial showing how to use 21CMMC with multinest, since it's a pretty big feature.

The `external_table_path` thing is something that I think has happened before. I'll try to look into it, but it might not be til next week unless I can find a spare moment.

src/py21cmmc/mcmc.py
qyx268: Then for input parameters that are not shared by emcee and multinest, I think I will need to put them into 
`mcmc_options={continue_sampling':True, 'reuse_burnin':True, 'log_level_21CMMC':None, 'sampler_cls':CosmoHammerSampler}` and 
`multinest_options={'log_level_21CMMC': None, 'n_live_points': 100, 'importance_nested_sampling': True, 'sampling_efficiency': 0.8, 'evidence_tolerance': 0.5, 'max_iter': 50, 'multimodal': True, 'write_output': True, 'resume': True}`

However, this will change some existing scripts for using the emcee sampler. Do you have other suggestions?

qyx268:
> @qyx268 OK thanks for the update! I say we merge it without the ability to do the full outputting within the likelihood, and create an issue for that right now, so we don't forget to add the feature later.

okay!

> You should also make a (small) tutorial showing how to use 21CMMC with multinest, since it's a pretty big feature.

yes, I will include a tutorial for this feature

> The `external_table_path` thing is something that I think has happened before. I'll try to look into it, but it might not be til next week unless I can find a spare moment.

thx!

src/py21cmmc/mcmc.py
steven-murray: No, I think all of the new multinest options can be part of the existing `**mcmc_options`, and you can give them defaults in the function body by doing something like

```python
n_live_points = mcmc_options.get('n_live_points', 100)
```

steven-murray: Also, 'resume' does the same thing as 'continue_sampling', so that's one less parameter.

qyx268: done. but we might need to consider renaming `run_mcmc` because multinest is not strictly speaking an mcmc.

qyx268:
I'm not sure if we want to add a test for this feature because we need to instead MultiNest and PyMultiNest to use this feature...

steven-murray:
I think we definitely want to test it -- just at a very basic level. Those packages should hopefully be installable fairly easily, right? We just have to write down the install steps in the travis.yaml.

src/py21cmmc/mcmc.py
steven-murray: I think it's close enough that no-one is going to care :-)

qyx268:
@steven-murray okay, they are included now and the test passes. Could you also help check the test and see if it makes sense? 

It's easy to include `pymultinest` in `setup.py`. Do you have any suggestions on building `MultiNest` when installing `py21cmmc`? You can take a look at the travis file for its installation.

src/py21cmmc/mcmc.py
```diff
@@ -172,19 +204,56 @@ def run_mcmc(
...
+                importance_nested_sampling=importance_nested_sampling,
+                multimodal=multimodal,
+                evidence_tolerance=evidence_tolerance,
+                sampling_efficiency=sampling_efficiency,
+                init_MPI=False,
+            )
+            return 1
+
+        except OSError:
+            raise ImportError("You also need to build MultiNest library")
```
steven-murray: Let's point the user to https://johannesbuchner.github.io/PyMultiNest/install.html#id4 here so they know how to install it.

steven-murray (approved):
OK, I think all of this is good. There seems to be an extra jupyter notebook in `src/py21cmmc/MultiNest.ipynb` that shouldn't be there, so please remove that.

The test errors that remain must be due to the `external_table_path`?

src/py21cmmc/mcmc.py
qyx268: ok

qyx268:
> The test errors that remain must be due to the `external_table_path`?

this is the error
image

I think it's either the `external_table_path` has been wrongly set to `/home/travis/.21cmfast/` (see `EXTERNALTABLES` in `src/py21cmfast/inputs.py`) or `External_tables` was not correctly copied into `$HOME/.21cmfast` during installation.

qyx268:
hmm... I'm not familiar with tox... After a few explorations, I manage to put the installation of MultiNest in tox.ini. But it seems that we cannot set LD_LIBRARY_PATH in tox.ini, which is required by MultiNest. Therefore, I have to put it in .travis.yml.

But in fact, MultiNest is not a python package. So we can actually put all of the steps in .travis.yml (I had one build here: https://travis-ci.com/github/21cmfast/21CMMC/jobs/352959474)

I'm not sure which one is the optimal way. Actually, do you know any other better or more commonly used setup? @steven-murray

steven-murray:
@qyx268 I'll try giving this a go tomorrow. Travis/tox can be a bit of a pain for complex stuff.

steven-murray:
@qyx I think the way you've done it is fine, and it looks like it works OK. This stuff is tricky, and I don't know of a necessarily "better" way to do it. I'm thinking perhaps we might just do away with tox, because I for one pretty much never test different python versions on my local system -- I let travis do that! I'm happy for you to merge this now if you're happy with it.

Thanks for the awesome work!

PR #32, opened 2020-01-28
LF likelihood bug fix

### LINKED ISSUE #24: [BUG] Luminosity Function Bugs+questions
Hey,

In line 950 of likelihood.py the def compute likelihood has 'return lnl' in the wrong indentation level. It is currently within the for loop, so it returns the lnl value for the first redshift only. 

Also within the lnl for the Luminosity function (line 946 of likelihood.py)', the lnl calculation returns a fiducially calculated luminosity (i think) rather than the data input one.
I'm giving it
'lnl += 0.5 *np.sum(
    (self.noise[i]["sigma"][:,1] - 10**model_spline(self.noise[i]["sigma"][:,0])) **2
     / self.noise[i]["sigma"][:,2] ** 2
)'
This should work given a cut for the magnitudes removing data for Muv<-20 has been made (I've just removed theses points from the input file (which is self.noise[redshift]["sigma"][Muv, phi, sigmaphi] format). 
I think (?) there should be a flag to tell it when to use data (like the line I provided above) or when to use the generated fiducial (like in line 946).

Finally I'm a bit confused by lines 571-575 in core.py  
Is this suppose to add some normal noise to the calculated LFs?
I get an error from the  ' scale = s(muv) ' argument  - I'm not sure what the standard deviation should be if adding normal noise onto these LFs?  
(Currently I'm just leaving this bit out).

Thanks,
Tom

binnietom:
Closes #24 

I think the redshift array and dictionary labels are the wrong way round.


steven-murray (commented):
OK. So, the `data` and `noise` attributes are definitely lists of dicts, rather than just dicts.
The reason that it is a list is because the base `LikelihoodBaseFile` class defines the data reading function to always return a list of dicts, because it might have multiple files. Also, the write function assumes the data to be a list of dicts. That means both the `_read_data` and `reduce_data` functions must return _the same_ kind of object -- a list of dicts. At the moment, they don't.

Having thought about it a bit more, I think there's a bit of a bigger problem here. I'm gonna have to think it through a bit harder.

steven-murray (approved):
These changes look good.

PR #36, opened 2020-02-17
add default LFs, taken from Park+18

### LINKED ISSUE #35: default LF data
**Is your feature request related to a problem? Please describe.**

To include default observational LF in the data folder 

**Describe the solution you'd like**

I take the data from Park+18 21CMMC and convert it into the required format. The data includes LF and noise at z=6, 7, 8 and 10.

**Describe alternatives you've considered**

**Additional context**

qyx268:
Closes #35


steven-murray:
Thanks @qyx268! I think in tandem with this we should update the tutorial to point to this data, so that users know where to find it and that they can use it. It might also be good to be able to pass a string to the LF likelihood which automatically finds this data.

qyx268:
Hi @steven-murray , I've modified the tutorial a bit and inform the users there are default LFs provided and also show them how to use LF core/LH for each redshift. I'm still a bit confused about the previous example where multiple redshifts are passed to the core function while nothing is given to the likelihood? In this case, how to compare with the observations if no datafile or noisefile is given. But if we specify those files, then we need to define core and likelihood for each redshift and pair them using the name. Then why would we provide this functionality to allow users to pass multiple redshifts? 

Let me know if u think more details should be provided. Also, I've added some lines to direct datafile and noisefile to the provided files and raise errors when users are asking for LFs at redshifts not in [6,7,8,10]

I'm also a bit confused about why self.redshifts is an array rather than just a number after paring... did you always put them into an array even when it's just a number?

qyx268:
I see, so the multi-redshift feature is for mock observation, right?

src/py21cmmc/likelihood.py
```diff
@@ -1054,14 +1058,37 @@ class LikelihoodLuminosityFunction(LikelihoodBaseFile):
 
     def __init__(self, *args, name="", **kwargs):
         super().__init__(*args, **kwargs)
-        if self.datafile is not None and len(self.datafile) != 1:
```
steven-murray, line 1052: I *think* we want to leave these statements here -- they also apply if self._simulate = True

src/py21cmmc/likelihood.py
```diff
@@ -1054,14 +1058,37 @@ class LikelihoodLuminosityFunction(LikelihoodBaseFile):
...
+                if self.redshifts[0] not in [6, 7, 8, 10]:
+                    raise ValueError(
+                        "only LFs at z=6,7,8 and 10 are provided! use your own LF :)"
+                    )
+                self.datafile = path.join(
+                    path.dirname(__file__),
+                    "data",
+                    "LF_lfuncs_z%d.npz" % self.redshifts[0],
+                )
+            else:
```
steven-murray: We can remove this if you keep in the above checks.

src/py21cmmc/likelihood.py
```diff
@@ -1054,14 +1058,37 @@ class LikelihoodLuminosityFunction(LikelihoodBaseFile):
...
+                        "can only pass a single datafile to LikelihoodLuminosityFunction!"
+                    )
+            if self.noisefile is None:
+                self.noisefile = path.join(
+                    path.dirname(__file__),
+                    "data",
+                    "LF_sigmas_z%d.npz" % self.redshifts[0],
+                )
+            else:
+                if len(self.noisefile) != 1:
```
steven-murray: Also can remove this if we keep the above checks

steven-murray (commented):
Looks great -- just check the checks on length of datafile to make sure they make sense.

docs/tutorials/mcmc_global.ipynb
```diff
@@ -340,18 +340,24 @@
    "outputs": [],
```
steven-murray: <p>I <em>think</em> you shouldn't have to specify the path to the data/noisefile here, right? Because they should be used by default if simulate=False and no files are provided?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/21cmfast/21CMMC/pull/36/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

steven-murray:
@qyx : the multi-redshift feature is for when you have a single data/noise file which contains multiple redshifts (i.e. it contains an array with shape `(Nz, Nuvbins)`). In this case, the redshifts in the file are assumed to be the same as those you passed as a list to the `core`. 

That's also why `redshifts` is always a list -- just in case multiple redshifts are passed in to do the above. 

However, it's much more flexible to make each file correspond to a different redshift. This way, each redshift can have a different number of bins specified.

qyx268:
I'm testing these data by running MCMC with them. I'll update once it's done.

qyx268:
The default data has been corrected imported. We can now use them by simply calling e.g.
`CoreLuminosityFunction( redshift=6, sigma=0, name='lfz6')`
and 
`LikelihoodLuminosityFunction(name='lfz6', simulate = False, )`

qyx268:
@steven-murray  I think the coverage decreased a bit because several lines in the likelihood.py file might look similar to the existing ones, in particular the raise ValueError line...

docs/tutorials/mcmc_global.ipynb
```diff
@@ -220,26 +220,14 @@
   },
```
steven-murray: <p>I think it would be useful to have some text here about the fact that there are default luminosity functions, but also how to use your own.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/21cmfast/21CMMC/pull/36/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

steven-murray (commented):
Almost there. I will check out the coverage issue -- can you add a bit more explanation into the notebook about how to use the LF likelihood (and that there are default luminosity functions)?

steven-murray:
With the coverage -- I think now is a good time to make sure we test some of these exceptions. WIth pytest you can use

```
with pytest.raises(ValueError):
   do_something()
```
to ensure that it raises an error. You could try adding a couple of tests that make sure that some errors are raised if you for example pass more than one redshift. This will get the coverage to go upwards.

Eventually, we want 95%+ coverage.

steven-murray (approved):

PR #38, opened 2020-05-22
fix: remove ability to load-balance by forcing good parameter choices…

### LINKED ISSUE #9: [BUG] Odd behaviour for zero likelihood
Hi, 

I am testing the MCMC "only" features i.e. as much as possible independently of 21cmFast.

I build a simple core and likelihood :

    class MyDummyCore(mcmc.CoreBase):
        def __init__(self, dummy,
                 *args, **kwargs ):
            super().__init__(*args, **kwargs)
            self.dummy = dummy
        
        def build_model_data(self, ctx):
            super().build_model_data(ctx)
            ctx.add("dummy", self.dummy)
        
    class MyDummyLikelihood(mcmc.LikelihoodBase):    
        def reduce_data(self, ctx): 
            params = ctx.getParams()
            return dict(M_TURN=params[0], ALPHA_STAR=params[1])
    
            ### ok
            #return -0.5* ( (model['M_TURN']-9)**2/0.25**2 + (model['ALPHA_STAR'])**2/0.125**2 )
        
            ### border effect ?
            #return -0.5* ( (model['M_TURN']-9)**2/1.**2 + (model['ALPHA_STAR'])**2/0.5**2 )
        
            ### cross ?
            return 0.
   
    core = MyDummyCore( 0. )
    likelihood = MyDummyLikelihood(  )

and a simple MCMC: 

    chain = mcmc.run_mcmc(
    core, likelihood,        # Use lists if multiple cores/likelihoods required. These will be eval'd in order.
    datadir='data_dummy',    # Directory for all outputs
    model_name=model_name,   # Filename of main chain output
    params=dict(             # Parameter dict as described above.
        M_TURN = [9.0, 6.0, 12.0, 4.0], # fid, rangeMin, rangeMax
        ALPHA_STAR = [0.0, -1.5, 1.5, 2.0],
    ), 
    walkersRatio=4,          # The number of walkers will be walkersRatio*nparams
    burninIterations=0,      # Number of iterations to save as burnin. Recommended to leave as zero.
    sampleIterations=10000,    # Number of iterations to sample, per walker.
    threadCount=16,           # Number of processes to use in MCMC (best as a factor of walkersRatio)
    continue_sampling=False  # Whether to contine sampling from previous run *up to* sampleIterations.
)

So the lnL is set to constant 0. I expect a flat distibution of the paramters in the end :
but there is a clear X shape... I do not think this is a border issue. I suspect a deeper problem in the mcmc...

MCMC_cross

then, in the likelihood function, if the distributions are wide gaussian, the result is boxy instead of perfectly round as expected:

MCMC_box

steven-murray:
… as it destroys load balancing.

Fixes #9 .

Proof that this is required: 

image

That's the LF likelihood only, which shouldn't constrain `alpha_esc` or `f_esc` at all. The blue has this feature remove(from @qyx268). 

It's not a feature of the initial positions either, as @BradGreig pointed out:
image

Here the initial positions are flat/gaussian ball for red/blue, but results in the same posterior (if it goes long enough).


qyx268 (approved):
looks good!

PR #39, opened 2020-05-27
21cm cmb

qyx268:
Several things we probably need to include before accepting this PR:

1. an automatic installation of `plik` and `class` if the users don't have them, or some kind of help... 
2. I've already added an option `use_21cmfast`. This is set to True by default, meaning we will get the EoR history from 21cmfast. But when `use_21cmfast=False`, we use the tangent function provided by class for the EoR where `z_reio` and `reionization_width` are given by `F_STAR10` and `ALPHA_STAR`. I'm doing this temporarily because it doesn't seem to make sense to include these two additional parameters in 21cmFAST considering it does not actually run 21cmFAST.
3. do we want to have `z_HeII` the same as `Zreion_HeII` in `Globals.h`? However, we hardly use `Zreion_HeII` in 21cmfast (except when calculating the CMB optical depth)


qyx268:
better to provide a tutorial of using this feature (e.g. the one presented in the cmb paper.)

src/21CMMC.egg-info/PKG-INFO
```diff
@@ -0,0 +1,86 @@
+Metadata-Version: 1.1
```
steven-murray: This file should not be included -- put it in the .gitignore (all of `21CMMC.egg-info`)

src/py21cmmc/core.py
```diff
@@ -645,3 +645,245 @@ def convert_model_to_mock(self, ctx):
             except TypeError:
 
                 lfunc[i] += np.random.normal(loc=0, scale=s, size=len(lfunc[i]))
+
+
+class CoreCMB(CoreBase):
+    r"""A Core Module that calls the CLASS CMB code and computes  Cl^TT,TE,EE and phiphi (the lensing potentials). It takes as an input the reionization history from 21cmFAST and a few cosmological parameters.
```
steven-murray: Split this line up -- the first line of the docstring is a one-sentence summary (see eg. https://developer.lsst.io/python/numpydoc.html for advice on formatting docstrings).

src/py21cmmc/core.py
```diff
@@ -645,3 +645,245 @@ def convert_model_to_mock(self, ctx):
...
+
+    Parameters
+    ----------
+    z_extrap_min : minimal z for reionization in CLASS. should basically always be set to 0.
```
steven-murray: These also are not formatted correctly (see above link, or other examples in the code)

src/py21cmmc/core.py
```diff
@@ -645,3 +645,245 @@ def convert_model_to_mock(self, ctx):
...
+        except AttributeError:
+            raise AttributeError(
+                "You must have compiled the classy.pyx file. Please go to "
+                + "/path/to/class/python and run the command\n "
+                + "python setup.py build"
+            )
+
+    def setup(self):
+        """Perform any post-init setup of the object."""
+        CoreBase.setup(self)
```
steven-murray: It's better to do `super().setup()` so if we ever change the name of `CoreBase` it will still work.

src/py21cmmc/core.py
```diff
@@ -645,3 +645,245 @@ def convert_model_to_mock(self, ctx):
...
+                print("import CLASS")
+            global cosmo
+            cosmo = Class()
+            self.verbose = verbose
+            self.z_extrap_min = z_extrap_min
+            self.z_extrap_max = z_extrap_max
+            self.z_HeI = z_HeI
+            self.z_HeII = z_HeII
+            self.use_21cmfast = use_21cmfast
+        except AttributeError:
```
steven-murray: I think this could also fail with an `ImportError`, which you'll need to catch.

src/py21cmmc/core.py
```diff
@@ -645,3 +645,245 @@ def convert_model_to_mock(self, ctx):
...
+                "spectra_verbose": self.verbose,
+                "nonlinear_verbose": self.verbose,
+                "lensing_verbose": self.verbose,
+            }
+
+        ##############
+        #
+        # call CLASS
+        #
+        ###############
```
steven-murray, line 1105: Make this just a single line comment

src/py21cmmc/core.py
```diff
@@ -645,3 +645,245 @@ def convert_model_to_mock(self, ctx):
...
+            }
+
+            self.io_options.update(io_options)
+
+        try:
+            from classy import Class
+
+            if verbose > 0:
+                print("import CLASS")
+            global cosmo
```
steven-murray, line 954: setting `cosmo` as a global parameter is almost certainly  a bad idea -- what's the justification for this?

src/py21cmmc/likelihood.py
```diff
@@ -682,6 +684,242 @@ def store(self, model, storage):
...
+    datafolder : folder that contains Planck data if "clik" format.
+    name_lkl : str, default = Planck_lowl_EE, the planck likelihood to compute. choice: Planck_lensing, Planck_highl_TTTEEE, Planck_lowl_EE
+    """
+
+    required_cores = ((core.CoreLightConeModule, core.CoreCMB),)
+
+    def __init__(
+        self,
+        *args,
+        datafolder="blah",
```
steven-murray: You don't need to give it a default -- "blah" is not a good default.

src/py21cmmc/likelihood.py
```diff
@@ -682,6 +684,242 @@ def store(self, model, storage):
...
+
+        Returns
+        -------
+        lnl : float
+            The log-likelihood for the given model.
+        """
+        cl = model["cl_cmb"]
+
+        if self.lensing:
+            my_clik = my_clik_lensing
```
steven-murray, line 766: This is super confusing because `my_clik_lensing` is a global. I think the reason for it being global is that it can't be pickled, so it can't be stored as part of the class. If this is the case, that's fine-ish (there are ways around it, but you then have to implemente `__getstate__` and `__setstate__` on the object, which is kind of annoying).  However, I'd at least add a comment like `# this is a global variable defined in initialize_clik_and_class`.

src/py21cmmc/likelihood.py
```diff
@@ -682,6 +684,242 @@ def store(self, model, storage):
...
+        #
+        # fill with Cl's
+        index = 0
+        if not self.lensing:
+            for i in range(length):
+                if my_clik.get_lmax()[i] > -1:
+                    for j in range(my_clik.get_lmax()[i] + 1):
+                        if i == 0:
+                            tot[index + j] = cl["tt"][j]
+                        if i == 1:
```
steven-murray: should be `elif`

src/py21cmmc/likelihood.py
```diff
@@ -693,20 +931,22 @@ class LikelihoodPlanck(LikelihoodBase):
...
-    # The Planck prior is modelled as a Gaussian: tau = 0.058 \pm 0.012
-    # (https://arxiv.org/abs/1605.03507)
-    tau_mean = 0.058
-    tau_sigma = 0.012
+    # The Planck prior is modelled as a Gaussian: tau = 0.058 \pm 0.012 (https://arxiv.org/abs/1605.03507) tau_mean = 0.058 tau_sigma = 0.012
+    # Cosmology from Planck 2018(https://arxiv.org/pdf/1807.06209.pdf)
+    # TT TE EE lowE lensing tau_mean = 0.0544 tau_sigma = 0.0073
+    # TT lowE
+    tau_mean = 0.0522
+    tau_sigma = 0.0080
```
steven-murray: I can't quite follow the progression of thought in these comments, and why `tau_mean` and `tau_sigma` have been changed.

src/py21cmmc/likelihood.py
```diff
@@ -792,7 +1032,7 @@ class LikelihoodNeutralFraction(LikelihoodBase):
     and 0 otherwise.
     """
 
-    required_cores = ((core.CoreLightConeModule, core.CoreCoevalModule),)
+    required_cores = ((core.CoreLightConeModule, core.CoreCoevalModule), (core.CoreCMB))
```
steven-murray: I think this will _enforce_ that `core.CoreCMB` is loaded, which is probably not what you want?

src/py21cmmc/likelihood.py
```diff
@@ -825,11 +1065,12 @@ def __init__(self, redshift=5.9, xHI=0.06, xHI_sigma=0.05):
         )  # these will become the redshifts of all coeval boxes, if that exists.
         self._use_coeval = True
         self._require_spline = False
+        self._use_tanh = False
 
     @property
     def lightcone_modules(self):
         """All lightcone core modules that are loaded."""
-        return [m for m in self._cores if isinstance(m, core.CoreLightConeModule)]
+        return [m for m in self._cores if (isinstance(m, core.CoreLightConeModule))]
```
steven-murray: this change is not required

steven-murray (commented):
I haven't yet looked at the tutorial, so I'll get to that later. I've provided a few comments here for some improvement. This is really cool, and will be a great addition. As for your three questions:

1. Yes, we should provide either automatic installation, or explicit instructions (I tried to get an auto installation of CLASS once and failed, but maybe you can :-))
2. Re-use as much of 21cmFAST parameters as you can, I think.

src/py21cmmc/likelihood.py
qyx268: This is a bit detailed about using clik. Do you mind commenting on this please @PoulinV ?

src/py21cmmc/likelihood.py
qyx268: Sorry, this was meant to be a temporary change since I was using a different set of the cosmological parameters in Planck2018. I will change it back to the original value. But do you think we can set tau_mean and tau_sigma as optional arguments?

src/py21cmmc/likelihood.py
qyx268: What I wanted to include is that we must have either one of these three options: `core.CoreLightConeModule`, `core.CoreCoevalModule`, `core.CoreCMB`. I can't remember why I put `core.CoreCMB` outside `(core.CoreLightConeModule, core.CoreCoevalModule)`. But why did we use `((core.CoreLightConeModule, core.CoreCoevalModule),)` rather than `(core.CoreLightConeModule, core.CoreCoevalModule)` in the first place?

PoulinV:
Hi guys!

I will look into all of this asap, I am just back from a week off from work to move into a new place and work om the house. Thanks for the hard work!! I see there was quite a lot of discussion about prior issues was that resolved? 

cheers
Vivian

steven-murray:
@PoulinV yes, the prior issue has been resolved :-)

src/py21cmmc/likelihood.py
steven-murray: I think this is what @PoulinV and I had discussed just after the Pisa workshop, and we came to the agreement that at least at first, the best way is to use it as a global. So that's fine, but it would be good to just have a comment over the variable telling us where it is defined.

src/py21cmmc/likelihood.py
steven-murray: Yes, we can set them as arguments in `__init__`, that will be better.

src/py21cmmc/likelihood.py
steven-murray: Having them inside a tuple inside the tuple means that you can have _either_ `CoreLightConeModule` or `CoreCoevalModule` to make the likelihood work. Having them like
`(core.CoreLightConeModule, core.CoreCoevalModule)` means you need _both_ to be loaded. Essentially, top-level of the tuple is AND, and second-level tuples are OR.

steven-murray:
@qyx268 let me know when you think it's ready for another review.

qyx268:
it seems that this PR is also experiencing the `external_table_path` issue happening in #27

src/py21cmmc/likelihood.py
```diff
@@ -682,31 +684,283 @@ def store(self, model, storage):
...
+    name_lkl : str
+        default = Planck_lowl_EE, the planck likelihood to compute. choice: Planck_lensing, Planck_highl_TTTEEE, Planck_lowl_EE
+    """
+
+    required_cores = ((core.CoreLightConeModule, core.CoreCMB),)
+
+    def __init__(
+        self,
+        *args,
+        datafolder="./",
```
qyx268: I'm wondering if we need to provide the Planck likelihood packages, which can be pretty large if we put them in the `data` folder (e.g. 80MB for Planck 2018 and >200MB for 2015). `datafolder` here refers to the folder where these likelihood codes are stored. when different `name_lkl` is used, users need to change `datafolder` as well to the folder where `name_lkl` is stored. The original Planck likelihood packages adopt folder names which are not very comprehensive (e.g.  `low_l/simall/simall_100x143_offlike5_EE_Aplanck_B.clik`)

I guess if we want to provide plik along with py21cmmc, we need to alter the folder names accordingly, so that it's easy for users who are not familiar with plik to use this likelihood

qyx268: However, without plik, I cannot add test for this likelihood...

steven-murray: I *think* it would be best to have an automatic downloader that gets the plik likelihoods when requested, and puts them somewhere in the user's home folder or something. This might make testing a bit annoying, but I think we can auto-cache it on travis as well.

qyx268:
hmm... it seems that the python in `tox` cannot access the packages that installed for the python in `travis virtualenv`...

qyx268:
it should be a way to install `clik` and `classy` (as shown in `travis.yml`) in `setup.py`, which is also needed for `MultiNest` in #27

steven-murray:
@qyx268 yes, the `tox` env is a separate virtual environment. You should do all that installation stuff in the `tox` env rather than in `travis.yml`. We use the `travis.yml` basically just to set up the env enough to have `tox`, then `tox` takes care of the rest.

This is a little convoluted, but it allows us to test in multiple environments locally as well.

qyx268:
> @qyx268 yes, the `tox` env is a separate virtual environment. You should do all that installation stuff in the `tox` env rather than in `travis.yml`. We use the `travis.yml` basically just to set up the env enough to have `tox`, then `tox` takes care of the rest.
> 
> This is a little convoluted, but it allows us to test in multiple environments locally as well.

Then I guess it's better to install those packages (plik, classy and also the likelihood data; also multinest in #27 ) in setup.py. In this way, users don't need to install those packages by themselves...

steven-murray:
@qyx268 yes, that is always preferable. But it seems like CLASS is not that easy to install, and it might be in our best interest just to make it optional, and point the user to their documentation to get it installed (otherwise we have to always upkeep how it gets installed on every OS, which can be a pain). 

This would mean having explicit installation in the tox env though. As for the plik likelihoods etc., I wonder if we shouldn't have a function in the code itself that will download them the first time they are used to somewhere on the system, and then after that, just use those files. You could have a look at astropy's `download_data` function.

Looks like the only test now failing is the planck one, which is saying that `classy` is not installed.

qyx268:
I managed to install clik and classy in tox environment. But I'm not sure why clik still cannot be found...

steven-murray:
I'll have a good look at this tomorrow!

steven-murray:
@qyx268 Well, I did something :man_shrugging: 

Now, all the C-codes (and likelihoods) are installed in the travis script, and they are cached, so that should make things faster at least. Also, to pass the environment variables along to `tox` (from the `clik_profile.sh`) required setting the `passenv` in `tox`. Finally, it required setting the `PYTHONPATH` to include both the default directory and the one from `clik`. 

After all that, it does seem to be finding the `clik` library, but it's still giving an error which I have no idea about. Reckon you could have a look?

steven-murray:
Also, would using [this package](https://pypi.org/project/classylss/) work instead of CLASS itself? It would be a lot easier to install..

qyx268:
I believe this crash is from CLASS. However, pytest passes locally. I'm worried the issue lies in the compatibility between CLASS with other libraries specified in the travis environment. I need to learn how to run travis locally otherwise debugging can be challenging when I need to make push to see the outcomes at every step...

PoulinV:
to help debug you can add some verbose to class, have you tried?
you need to add the following flags when setting all the parameters in CLASS
‘input_verbose': 1,
‘background_verbose' : 1,
‘thermodynamics_verbose': 1,
‘perturbations_verbose’: 1,
‘primordial_verbose’: 1,
‘spectra_verbose’: 1,
‘nonlinear_verbose’: 1,
‘lensing_verbose’: 1,
‘output_verbose’: 1

qyx268:
Are you able to get results with this EoR? My program just stuck at the last line. 

```
import classy
cosmo = classy.Class()

redshift_class = np.array([
        0.        ,  3.        ,  4.        ,  5.        ,  5.6       ,
        6.26      ,  6.986     ,  7.7846    ,  8.66306   ,  9.629366  ,
       10.6923026 , 11.86153286, 13.14768615, 14.56245476, 16.11870024,
       17.83057026, 19.71362729, 21.78499001, 22.        ])

xe = np.array([
       -2.00000000e+00, -2.00000000e+00, -1.00000000e+00,  9.94080095e-01,
        9.69432181e-01,  9.19968776e-01,  7.59219393e-01,  4.66289520e-01,
        2.90094316e-01,  1.71188235e-01,  9.50848460e-02,  4.89555597e-02,
        2.11643577e-02,  7.73453712e-03,  2.41136551e-03,  6.00934029e-04,
        0.00000000e+00,  0.00000000e+00,  0.00000000e+00])

common_settings = {
                "output": "tCl, pCl, lCl",
                "lensing": "yes",
                "l_max_scalars": 3000,
                "reio_parametrization": "reio_inter",
                "reio_inter_num": len(xe),
                "reio_inter_z": ",".join(
                    ["%.5f" % x for x in redshift_class]
                ),
                "reio_inter_xe": ",".join(["%.5e" % x for x in xe]),
                'input_verbose': 1,
                'background_verbose' : 1,
                'thermodynamics_verbose': 1,
                'perturbations_verbose': 1,
                'primordial_verbose': 1,
                'spectra_verbose': 1,
                'nonlinear_verbose': 1,
                'lensing_verbose': 1,
                'output_verbose': 1
}
cosmo.set(common_settings)
cosmo.compute()
```

qyx268:
Maybe I haven't done the setting correctly. I did remove some lines when I clean the scripts. It would be great if you can review the code? @PoulinV

PoulinV:
Yes I think the problem is you are passing negative x_e to CLASS.
The trick of replacing negative xe by the value after the first and second Helium reionization, I believe we implemented in CoreCMB?

qyx268:
```
            # Translate xHI into xe for CLASS.
            # The option -1, -2 ensure helium first and second reionization respectively at z_HeI and z_HeII.
            xe = 1 - xHI
            xe = np.concatenate(([-2, -2, -1], xe, [0]))
            redshift_class = np.concatenate(([0, z_HeII, z_HeI], redshifts, [z_xe_0]))
```

I thought this is what we should be doing? giving -1 and -2 for helium reionization

PoulinV:
sorry you are right, CLASS does understand these flags.

but surely something goes wrong when interpolating, not sure what. I will test it again,

qyx268:
Thanks. So you can also reproduce this error?  Then we might need to raise it to the classy repository too.

steven-murray:
@qyx268 there are ways you can get travis to run locally using Docker -- might be worth giving it a go.

qyx268:
thx @steven-murray , that's what I tried today. I can confirm that for some EoR histories, CLASS freezes and just accumulates RAM. I suspect this error in travis is caused by CLASS rather than 21CMMC. @PoulinV is looking into this.

steven-murray:
Nice sleuth work!

qyx268:
... ... why has this sunk? @PoulinV  How's the communication with the CLASS team?

steven-murray:
@qyx268 and @PoulinV  how is this going?

qyx268:
It seems that we cannot get support from the CLASS team. I suggest we stop trying to make it to work in the test. Then I can merge this PR.

steven-murray:
Sounds like a good plan. You can mark the test as `@pytest.mark.xfail` so that we can keep it, but expect it to fail until we get support. That'll remind us to do it at some point...

qyx268:
one thing that i'm not sure if that I had to remove the requirements_dev.txt file to pass pre-commit in Github Action, which was not an issue locally. But I think we do not need this file, do we? There is an environment_dev.yml file which serves the purpose? @steven-murray

qyx268:
I also pytest.mark.xfail-ed test_planckpowerspectra and updated according to master. This PR should pass now and hopefully can be approved? @steven-murray

qyx268:
macos-latest 3.7 failed because of an installation/compiling erro for h5py...
macos-latest 3.8 failed by test_continue_burnin because diff two arrays return False, which by eye are the same... 

I just rerun the job, hopefully these would be gone...

qyx268:
hmm... looks like the failed tests all passed now. @steven-murray do you know what is happening to tcodecov? This seems to keep happening that codecov is not reporting

steven-murray:
@qyx268 , yes removing the `requirements_dev.txt` is perfectly fine. The codecov thing has been a problem in all my repos. I've updated the action so that hopefully it works now.

setup.py
```diff
@@ -70,8 +70,19 @@ def _find_version(*file_paths):
         "emcee<3",
         "powerbox>=0.5.7",
         "cached_property",
-        "21cmFAST>=3.0.0dev1",
+        "21cmFAST>=3.0.0dev6",
```
steven-murray, line 73: Should update this to be `21cmFAST>=3.0<4.0`

steven-murray (commented):
Thanks @qyx268 -- I'm close to approving, however there are a couple of outstanding comments (probably resolvable just by discussion, rather than changing code).

There are also a few files that may not be intended to be included -- eg. the `docs/tutorials/warmup.py` and the `docs/tutorials/test_class_21cmfast.ipynb`?

setup.py
qyx268: okay!

qyx268:
Hey @steven-murray , I think those files were added by @PoulinV to help me understand how CLASS works. But I agree, they don't need to be there and I've removed them.

qyx268:
image

this happened again but now for macos-latest 3.8.

steven-murray (approved):
Thanks @qyx268, happy for it to be merged now! Looks like whatever error that was has now gone away?

steven-murray:
Ah, I forgot! Please update the changelog!

qyx268:
@steven-murray  yeah, I rerun the test twice, and it's gone...  The changelog was updated. But do you want to wait to see if codecov reports anything in the end? or we can just merge as admin.

steven-murray:
eh. I know how to fix it.... gimme a sec.

steven-murray:
Nope, I have no idea why it's not working. It's giving an error that the files weren't able to be processed (on Codecov). But doesn't say why...

steven-murray:
I'm just gonna go ahead and merge. A lot of the infrastructure for this repo has to be updated anyway.

PR #41, opened 2020-06-23
Fix external tables

steven-murray:
Updates the dep of 21cmFAST to dev6 -- dev3-5 were missing external tables in their built distributions.


steven-murray:
@qyx268 this is the fix for both of the other PRs. Feel free to approve and merge, then merge master into the other two PR branches. In fact, they should already be working, even without this PR (because the fix was really in 21cmFAST).

qyx268 (approved):

PR #43, opened 2020-07-07
fix: cache full coeval/lightcone as initialization

### LINKED ISSUE #42: [BUG] PF being run every iteration
**Describe the bug:**

@qyx268 and @catherinewatkinson both report that `perturb_field` is being run every iteration, despite none of the params being cosmological. This would be horrible.

**To Reproduce:**
I will let @catherinewatkinson and/or @qyx268 post steps to reproduce.

**Expected behavior:**

If no cosmological parameters are being changed (only `AstroParams`), then only `spin_temp` and/or `ionize_box` should be re-run on any iteration. The init and perturb field should be computed and stored on disk at the beginning, and read in on every iteration.

steven-murray:
Fixes #42 

This aims to fix the caching of init and perturb boxes by performing a full saved run of the coeval/lightcone default parameters as part of `setup()`.

The saved coeval/lightcone can also be used as part of post-MCMC analysis.


steven-murray:
@qyx268 can I get you to run your test script over it quickly to see if it exhibits the right behaviour?

qyx268 (approved):
all good. tests show perturbs are only called once at the beginning. all fields including perturb, ionized, spin and brightness are stored for the reference model.

PR #45, opened 2020-08-05
fix: don't pass pool twice to sampler

steven-murray:
This fixes an issue where `pool` is passed twice to construct the sampler object.


BradGreig (approved):

steven-murray:
@qyx268 could you have a look at why this won't install on macosx for python 3.9+?

qyx268:
Hey @steven-murray , I made some ugly updates to get around the issues. Please take a look and see if there is a better way :)

qyx268:
@steven-murray  I have no idea why it takes more than 30min to install 21cmfast now...

PR #49, opened 2021-02-28
Lyalpha

qyx268:
This should be included in the master now since we've submitted the paper


src/py21cmmc/likelihood.py
```diff
@@ -185,7 +185,11 @@ def _read_data(self):
                     )
                 )
             else:
-                data.append(dict(np.load(fl, allow_pickle=True)))
+                try:
+                    data.append(dict(np.load(fl, allow_pickle=True)))
+                except TypeError:
+                    # TODO: need to better handle this
+                    data.append(dict(np.load(fl, allow_pickle=True).item()))
```
steven-murray: what's happening here?

steven-murray (approved):
This looks good. Just one comment to address, and I also noticed there were a bunch of "empty" files... is that expected?

src/py21cmmc/likelihood.py
qyx268: hmm... I think it was a numpy version thing that my forest data has to have .item() when load in

qyx268 (commented):
the npy files are not empty. They are just not readable by github I think

qyx268:
hmm.. why is travis not checking?

qyx268:
```Builds have been temporarily disabled for public repositories due to a negative credit balance. Please go to the Plan page to replenish your credit balance or alter your Consume paid credits for OSS setting.```

Is this why you moved 21cmfast away from travis? @steven-murray

**Comment by steven-murray** [comment]
Yes indeed it is. Also, Actions are just a bit nicer and more flexible.

Let me know if you want me to do this. I can do it tomorrow.

**Comment by qyx268** [comment]
I've copied/pasted from 21cmfast and updated it accordingly. Could you please take a look and update them if needed? Also, How do we stop requiring Travis test? It seems that we need to send an email to stop the service?

File: src/py21cmmc/core.py [hunk 2]
```diff
@@ -639,3 +651,214 @@ def convert_model_to_mock(self, ctx):
...
+class CoreForest(CoreLightConeModule):
+    r"""A Core Module that produces model effective optical depth at a range of redshifts.
+
+    name : str
+        The name used to match the likelihood
+
+    observation : str
+        The observation that is used to construct the tau_eff statisctic.
+
+    N_realization : int
```
steven-murray: Can we make this `n_realization` (all lower case)?

src/py21cmmc/core.py
```diff
@@ -639,3 +651,214 @@ def convert_model_to_mock(self, ctx):
...
+            or self.observation == "bosman_pessimistic"
+        ):
+            data = np.load(
+                path.join(path.dirname(__file__), "data/Forests/Bosman18/data.npy"),
+                allow_pickle=True,
+            ).item()
+            targets = (data["zs"] > self.redshift[0] - 0.1) * (
+                data["zs"] <= self.redshift[0] + 0.1
+            )
+            self.Nlos = sum(targets)
```
steven-murray: similarly here

src/py21cmmc/core.py
```diff
@@ -639,3 +651,214 @@ def convert_model_to_mock(self, ctx):
...
+            raise ValueError(
+                "You asked for %d realizations, larger than what the box has (Total los / needed los = %d / %d)! Increase HII_DIM!"
+                % (self.N_realization, self.user_params.HII_DIM ** 2, self.Nlos)
+            )
+
+    def setup(self):
+        """Run post-init setup."""
+        CoreBase.setup(self)
+
+    def tau_GP(self, Gamma_bg, Delta, Temp, redshifts):
```
steven-murray: again, here would like all variables to be lower case

src/py21cmmc/likelihood.py
steven-murray: are you sure it shouldn't be `.items()`? Is there a test of this?

steven-murray (commented):
Just a few stylistic comments, plus a request for a test specifically covering this likelihood (if possible!)

steven-murray:
I've setup the coveralls so that it actually reports a status, and also deactivated the travis checks.

qyx268:
I've updated the variable names. 

For the `.item` issue, maybe I didn't configure `data/Forests/Bosman18/data.npy` correctly? This is what happened if I did `.items()` or don't do `.item()`

image

qyx268:
@steven-murray  can i ask for some help? I don't think I'm understaning the test correctly. 

So I had to recreate a lc object and context because the forest data needs to reach z<6. I think lc_core_lowz_ctx should have the forest core in the context as I've added it into the cores when building the chain. However, in the test function, it seems that the ctx actually didn't run any simulation(?). Then I get an error reporting tau_eff is None.

How do I make sure the forest core is called and the modelled forest data is generated?

steven-murray:
Hi @qyx268. I'll give it a look -- I can't answer the question off the top of my head. Hopefully can get to it today, but I am pretty packed :/

steven-murray:
Hi @qyx268. Firstly concerning the numpy load. Can you try this:

image

I can't reproduce your errors, but I'll bet it's how you saved the data.

steven-murray:
I updated the test. I think the problem was that you were using the other `Core` to build the model data. You can use the `LikelihoodComputationChain` itself to build the model data, which will use all the available cores.

qyx268:
@steven-murray  thx for the help. For the coverall test, is it just because we didn't have one on master to compare against?

qyx268:
it's actually 81% coverage for the master...  google saied i need to close and reopen

PR #62, opened 2022-05-19
reference ParameterError in multinest likelihood

### LINKED ISSUE #61: [BUG] pymultinest: strange likelihood values for failed 21cmFAST parameters
**Describe the bug:**

Running 21CMMC with `pymultinest` makes a weird output for the failed 21cmFAST parameter values. Instead of `-inf`, it saves a small positive number for the likelihood, which makes the run completely biased to that region. 

I'm attaching a sample output for the [post_equal_weight](https://github.com/21cmfast/21CMMC/files/8686289/post_equal_weight.txt) file. The items with likelihood `0.467...-309` have parameter values equal to the failed 21cmFAST runs.

The same run with `emcee` behaves normally.

**Expected behavior:**
For failed runs a large negative log-likelihood values should be outputted. I suspect this is the problem of how `pymultinest` interprets `np.inf`. Maybe just a large float number would suffice.

dprelogo:
Due to not calling `LikelihoodComputationChain` `__call__` function directly, `ParameterError` is not catched. This makes MultiNest return `NaNs` instead of `-np.inf`.
`ParameterError` catch is thus added to the likelihood function directly.

### DIFF
--- a/src/py21cmmc/mcmc.py
+++ b/src/py21cmmc/mcmc.py
@@ -6,7 +6,6 @@ from os import mkdir, path
 from py21cmfast import yaml
 from py21cmfast._utils import ParameterError
 
-
 from .cosmoHammer import (
     CosmoHammerSampler,
     HDFStorageUtil,
@@ -211,22 +210,18 @@ Likelihood {} was defined to re-simulate data/noise, but this is incompatible wi
     if use_multinest:
 
         def likelihood(p, ndim, nparams):
-            inp = [
-                params[i][1] + p[i] * (params[i][2] - params[i][1]) for i in range(ndim)
-            ]
             try:
                 return chain.computeLikelihoods(
                     chain.build_model_data(
-                        Params(*[(k, v) for k, v in zip(params.keys, inp)])
+                        Params(*[(k, v) for k, v in zip(params.keys, p)])
                     )
                 )
             except ParameterError:
                 return -np.inf
 
         def prior(p, ndim, nparams):
-            p = [
-                params[i][1] + p[i] * (params[i][2] - params[i][1]) for i in range(ndim)
-            ]
+            for i in range(ndim):
+                p[i] = params[i][1] + p[i] * (params[i][2] - params[i][1])
 
         try:
             sampler = run(
--- a/src/py21cmmc/mcmc.py
+++ b/src/py21cmmc/mcmc.py
@@ -1,8 +1,11 @@
 """High-level functions for running MCMC chains."""
 import logging
+import numpy as np
 from concurrent.futures import ProcessPoolExecutor
 from os import mkdir, path
 from py21cmfast import yaml
+from py21cmfast._utils import ParameterError
+
 
 from .cosmoHammer import (
     CosmoHammerSampler,
@@ -211,11 +214,14 @@ Likelihood {} was defined to re-simulate data/noise, but this is incompatible wi
             inp = [
                 params[i][1] + p[i] * (params[i][2] - params[i][1]) for i in range(ndim)
             ]
-            return chain.computeLikelihoods(
-                chain.build_model_data(
-                    Params(*[(k, v) for k, v in zip(params.keys, inp)])
+            try:
+                return chain.computeLikelihoods(
+                    chain.build_model_data(
+                        Params(*[(k, v) for k, v in zip(params.keys, inp)])
+                    )
                 )
-            )
+            except ParameterError:
+                return -np.inf
 
         def prior(p, ndim, nparams):
             p = [
--- a/src/py21cmmc/mcmc.py
+++ b/src/py21cmmc/mcmc.py
@@ -6,7 +6,6 @@ from os import mkdir, path
 from py21cmfast import yaml
 from py21cmfast._utils import ParameterError
 
-
 from .cosmoHammer import (
     CosmoHammerSampler,
     HDFStorageUtil,
--- a/tests/test_multinest.py
+++ b/tests/test_multinest.py
@@ -1,7 +1,102 @@
+import pytest
+
+import numpy as np
+import pymultinest
+from functools import partial
+from py21cmfast._utils import ParameterError
+
 import py21cmmc as mcmc
+from py21cmmc.cosmoHammer import Params
+from py21cmmc.likelihood import LikelihoodBase
 
 
-def test_multinest():
+@pytest.fixture(scope="module")
+def astro_params():
+    ap = {
+        "L_X": [40.0, 38.0, 42.0, 0.05],
+        "NU_X_THRESH": [500.0, 200.0, 1500.0, 20.0],
+    }
+    return Params(*[(k, v) for k, v in ap.items()])
+
+
+@pytest.fixture(scope="module")
+def prior(astro_params):
+    class PriorFunction(LikelihoodBase):
+        def __init__(self, arg_names, f):
+            super().__init__()
+            self.arg_names = arg_names
+            self.f = f
+
+        def computeLikelihood(self, arg_values):
+            return self.f(arg_values)
+
+        def reduce_data(self, ctx):
+            params = ctx.getParams()
+            arg_values = [v for k, v in params.items() if k in self.arg_names]
+            return arg_values
+
+    def f(x, astro_params):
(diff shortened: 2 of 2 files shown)


steven-murray:
@dprelogo this looks good. My only question is whether we can't just call the chain directly instead. I presume we do this because we're scaling MultiNest's unit-hypercube prior out to our real-space prior. Firstly, we have to be careful with that -- if we have Gaussian priors or some other distribution other than uniform, this will break. Secondly, why don't we just scale the parameters out, then call the chain directly?

qyx268:
yes, that's right. It is all because multinest samples in unit hypercube, so I need to rescale them back to run the model and also I use this `run` function from `pymultinest` which has its own way of taking prior and likelihood as inputs.

steven-murray:
@qyx268 could the fact that the prior isn't returning anything be a problem?

qyx268:
For the concern on non-uniform priors, I think it is fine. I normally define my own prior (e.g. a Gaussian one) together when I define other likelihoods and then I just pass the prior to the list of likelihood. e.g. 
```
class Prior(LikelihoodBase):
    def reduce_data(self, ctx):
        params = ctx.getParams()
        return  [params.log10_f_rescale, params.f_rescale_slope]

        return -0.5 * ((model[0] / 5)**2 + (model[1] / 2.5)**2)

...
likelihood = [ LikelihoodPlanck(), Prior()]
```
Is there another way that I'm not aware of ?

qyx268:
> 

could you explain why this might happen?

qyx268:
> @qyx268 could the fact that the prior isn't returning anything be a problem?

Maybe the name `prior` is a bit confusing. But I followed the instruction from `pymultinest`, see here: https://github.com/JohannesBuchner/PyMultiNest/blob/e2f2e75bdb53d2f8dbd48874744cf4dd50e986e1/pymultinest/run.py#L91

qyx268:
@steven-murray  regarding to add a test for `ParameterError`, I don't think there is an easy way since we need to at least find a parameter set that triggers `ParameterError` in 21cmfast... if we decide to just go with try and except, perhaps, we need to lower the codecov requirement?

dprelogo:
> @dprelogo this looks good. My only question is whether we can't just call the chain directly instead. I presume we do this because we're scaling MultiNest's unit-hypercube prior out to our real-space prior. Firstly, we have to be careful with that -- if we have Gaussian priors or some other distribution other than uniform, this will break. Secondly, why don't we just scale the parameters out, then call the chain directly?

As I wrote in the bug thread, this was my main idea (to call `LikelihoodComputationChain` directly), however there is some kind of overhead (with respect to the current code) that I don't understand... both of them however solve the issue.

dprelogo:
> For the concern on non-uniform priors, I think it is fine. I normally define my own prior (e.g. a Gaussian one) together when I define other likelihoods and then I just pass the prior to the list of likelihood. e.g.
> 
> ```
> class Prior(LikelihoodBase):
>     def reduce_data(self, ctx):
>         params = ctx.getParams()
>         return  [params.log10_f_rescale, params.f_rescale_slope]
> 
>     def computeLikelihood(self, model):
>         return -0.5 * ((model[0] / 5)**2 + (model[1] / 2.5)**2)
> 
> ...
> likelihood = [ LikelihoodPlanck(), Prior()]
> ```
> 
> Is there another way that I'm not aware of ?

In principle it should be possible, but it seems non-trivial to implement... the result in both ways (passing it to the MultiNest directly or adding it as a likelihood) is the same... however, the code should converge faster with a more constraining prior directly passed to the MultiNest

steven-murray:
@qyx268, you're _almost_ following the instructions for PyMultiNest. It looks like the interface is _very_ similar to polychord. So, what they have as an example is to replace the input parameters with new parameters that are drawn from the distribution rather than the unit hypercube, so eg: 
image

You're doing this: 
image

Your code does nothing at all -- `p` is set as a new reference. You have to update its contents (like the multinest example) instead of defining `p` again in the function.

Here's what I do in my `yabf` code (which is supposed to be a better `cosmoHammer`):

image

So, I define a prior that takes the hypercube parameters, and returns a list of new parameters which get defined on based on the `.ppf` of a `scipy.stats` distribution.

qyx268:
You are right, by doing 'p=[ BLABLA ]' instead of the `for` loop, I've lost access to the original `p` and therefore the RAM that stores the original `p` doesn't get updated. 
image

But I remember when writing this piece of code, I compared the results between with and without the rescaling. If the results were the same, I wouldn't allow this to be included. Also, if what I've put in the `prior` function didn't work, I would have been always using the unit hypercube, which can't be the case given all inference results we have produced, e.g., the HERA one where we even have parameters between 38 and 42 (L_X) or 100 and 1500 (E_0). So I believe this is not how this python wrapper of multinest works.

steven-murray:
@qyx268 But I think you're doing this "scaling" again in your likelihood call, which is why it's using the right parameters in the end. But that's both unnecessary work, and also doesn't let MultiNest do anything appropriate with the prior (not sure it actually uses it in any specific way, so maybe that's OK).

steven-murray (commented):
@dprelogo I suggest you try and update the prior function to do the scaling, and then remove the `inp` part of the likelihood, so the likelihood should then just be `chain.computeLikelihoods(chain.build_model_data(params))` with the try/except.

dprelogo:
@qyx268 @steven-murray Does this look okay? Should we also support some other prior transformations? At least in the simplest Gaussian case this should be easy to implement, something with `.ppf` as in your example @steven-murray

steven-murray (approved):
Thanks @dprelogo, this looks correct to me. Can you confirm that it's working as expected? Is there a simple test that can be written?

dprelogo:
> 

I can confirm that it works correctly. However, the only difference in the flat-prior case is that the output files contain the actual parameter values, instead of 0-1. I wrote a simple test to check the output files if ranges are correct and if `ParameterError` works as intended.

steven-murray:
Well then, that sounds superior :-)

qyx268:
I got so confused as in the old 21CMMC, this is what I have 

```
    def likelihood(p, ndim, nparams):
        return chain(p)[0][0]
    def prior(p, ndim, nparams):
        p = [ params[i][0] + p[i]*(params[i][1]-params[i][0]) for i in range(ndim) ]
```
There is no scaling in the likelihood here. But then I found the following in `CosmoHammer_21CMMC/likelihood/module/Likelihood21cmFast.py `

```
    def Likelihood(self,ctx):

        if self.FlagOptions['USE_NEST'] is True:
            params_unity = ctx.getParams()
            params = [ self.params_range[i][0] + params_unity[i]*(self.params_range[i][1]-self.params_range[i][0]) for i in range(self.nim) ]
        else:
            params = ctx.getParams()

```
So yeah, I must add the rescaling here because after "almost" following the pymultinest instruction, I couldn't get correct results. 

I also did a quick test with 3 runs having likelihood and prior written as the following three:
```
        def likelihood(p, ndim, nparams):
            inp = [
                params[i][1] + p[i] * (params[i][2] - params[i][1]) for i in range(ndim)
            ]
                return chain.computeLikelihoods(
                    chain.build_model_data(
                        Params(*[(k, v) for k, v in zip(params.keys, inp)])
                    )
                )

        def prior(p, ndim, nparams):
            p = [
                params[i][1] + p[i] * (params[i][2] - params[i][1]) for i in range(ndim)
            ]
```
,
```
        def likelihood(p, ndim, nparams):
            inp = p
                return chain.computeLikelihoods(
                    chain.build_model_data(
                        Params(*[(k, v) for k, v in zip(params.keys, inp)])
                    )
                )

        def prior(p, ndim, nparams):
            p = [
                params[i][1] + p[i] * (params[i][2] - params[i][1]) for i in range(ndim)
            ]
```
and
```
        def likelihood(p, ndim, nparams):
            inp = p
                return chain.computeLikelihoods(
                    chain.build_model_data(
                        Params(*[(k, v) for k, v in zip(params.keys, inp)])
                    )
                )

        def prior(p, ndim, nparams):
            for i in range(ndim):
                p[i] = params[i][1] + p[i] * (params[i][2] - params[i][1]) 
```
Both the first (what we had) and 3rd (what @dprelogo has in this PR) give correct results.

qyx268 (approved):

PR #67, opened 2022-09-08
add minihalo for lf and remove print in core cmb

qyx268:


steven-murray (commented):
Thanks @qyx268, this looks good to me. Please update the changelog, then I'll approve.

steven-murray (approved):

PR #69, opened 2022-10-17
Slice Sampler for 21CMMC

samgagnon:
Implemented slice sampling for 21CMMC using the zeus package

### DIFF
--- a/src/py21cmmc/mcmc.py
+++ b/src/py21cmmc/mcmc.py
@@ -167,8 +167,8 @@ def run_mcmc(
 
     if use_zeus:
         ndim = mcmc_options.get("ndim", 2) # Number of parameters/dimensions (e.g. m and c)
-        nwalkers = mcmc_options.get(nwalkers, 10) # Number of walkers to use. It should be at least twice the number of dimensions.
-        nsteps = mcmc_options.get(nsteps, 100) # Number of steps/iterations.
+        nwalkers = mcmc_options.get("nwalkers", 10) # Number of walkers to use. It should be at least twice the number of dimensions.
+        nsteps = mcmc_options.get("nsteps", 100) # Number of steps/iterations.
         start = 0.01 * np.random.randn(nwalkers, ndim) # Initial positions of the walkers.
         try:
             import zeus
@@ -269,6 +269,11 @@ def run_mcmc(
             )
 
     elif use_zeus:
+        
+        def prior(p, ndim):
+            for i in range(ndim):
+                p[i] = params[i][1] + p[i] * (params[i][2] - params[i][1])
+            return p
 
         def likelihood(p):
             try:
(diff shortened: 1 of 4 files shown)


src/py21cmmc/mcmc.py
```diff
@@ -154,6 +200,47 @@ def run_mcmc(
...
         pass
 
+    if use_zeus:
+        ndim = mcmc_options.get(
+            "ndim", 2
+        )  # Number of parameters/dimensions (e.g. m and c)
+        nwalkers = mcmc_options.get(
+            "nwalkers", 10
+        )  # Number of walkers to use. It should be at least twice the number of dimensions.
+        nsteps = mcmc_options.get("nsteps", 100)  # Number of steps/iterations.
```
steven-murray, line 210: Can you list the default for nsteps in the docstring?

src/py21cmmc/mcmc.py
```diff
@@ -154,6 +200,47 @@ def run_mcmc(
     except FileExistsError:
         pass
 
+    if use_zeus:
+        ndim = mcmc_options.get(
+            "ndim", 2
+        )  # Number of parameters/dimensions (e.g. m and c)
+        nwalkers = mcmc_options.get(
+            "nwalkers", 10
```
steven-murray, line 208: Can you list the default for nwalkers in the docstring?

steven-murray: Also, could you make the default exactly 2*nparams?

src/py21cmmc/mcmc.py
```diff
@@ -183,9 +270,9 @@ def run_mcmc(
             if hasattr(lk, "_simulate") and lk._simulate:
                 logger.warning(
                     """
-Likelihood {} was defined to re-simulate data/noise, but this is incompatible with
-`continue_sampling`. Setting simulate=False and continuing...
-"""
+                    Likelihood {} was defined to re-simulate data/noise, but this is incompatible with
+                    `continue_sampling`. Setting simulate=False and continuing...
+                    """
```
steven-murray: This was my bad, but I think this would be better as single " quotes, one pair for each line, i.e.:

```python
logger.warning(
    f"Likelihood {lk} was defined to re-simulate data/noise, but this is incompatible with"
    "`continue_sampling`. Setting simulate=False and continuing..."
)
```

steven-murray (commented):
Thank you @samgagnon, this is really nice. I just had a few comments, and then two more bigger ones: (i) add a note to the readme, and (ii) write a test! Obviously the test will need to be very small, but it would be nice to know that the right inputs/outputs are being served up.

samgagnon:
Hi Steve, thanks for all the useful comments! I've addressed these all on my zeus fork. I think it should be safe to merge now.

steven-murray:
Ah, looks like you'll need to add zeus to the `test` requirements.

samgagnon:
ok, I just added `zeus` to `test-env.yml`. This should work now

steven-murray:
@samgagnon is `zeus` a conda package? You might have to add it to the `pip` section of the test env.

samgagnon:
The package is actually called `zeus-mcmc`, which has been amended in the test env

steven-murray:
@samgagnon looks like it's almost there. 

There are a couple of pre-commit failures that you'll need to fix (you should install pre-commit on your own system to get these errors on `git commit` rather than seeing them up here -- `pip install pre-commit; pre-commit install`). 

I don't know why the code coverage isn't reporting on linux. Codecov is just frustrating sometimes. I would just try the run again with the fix and hope it works. If not, we can just merge, and I'll try fixing it on another branch.

samgagnon:
Running `pre-commit` raises no errors locally. The codecov failure seem to be due to missing an access token of some type, and I'm not sure how to add such a token to my pull request. Do you think it's important, or is it ok to merge at this point?

steven-murray:
Let's merge, and I'll try to fix things on my end.

PR #72, opened 2023-02-15
Implemented HERA PS likelihood and emulator support

DanielaBreitman:


src/py21cmmc/core.py
```diff
@@ -1151,3 +1151,120 @@ def get_cl(self, cosmo, l_max=-1):
...
+        """Compute all data defined by this core and add it to the context."""
+        # Update parameters
+        logger.debug(f"Updating parameters: {ctx.getParams()}")
+        astro_params, cosmo_params = self._update_params(ctx.getParams())
+        logger.debug(f"AstroParams: {astro_params}")
+        logger.debug(f"CosmoParams: {cosmo_params}")
+        ## TODO check that cosmo_params = default params (wtv Yuxiang used) and if they are not, return err msg
+
+        # Call 21cmEMU wrapper which returns a dict
+        all_summaries = p21cmEMU(emu_path=self.emu_path,
```
steven-murray: it looks like you could initialize the `p21cmEMU` object once, and then call it for your astro params here?

src/py21cmmc/emulator.py
```diff
@@ -0,0 +1,197 @@
...
+import py21cmfast as p21
+import logging
+
+# Logging Config
+LOGGING_CONFIG = {}
+logging_format = "[%(asctime)s] %(process)d-%(levelname)s "
+logging_format += "%(module)s::%(funcName)s():l%(lineno)d: "
+logging_format += "%(message)s"
+logging.basicConfig(format=logging_format, level=logging.INFO)
+log = logging.getLogger()
```
steven-murray: don't put logging config into a module -- it breaks the setting of config in scripts that call the module.

src/py21cmmc/emulator.py
```diff
@@ -0,0 +1,197 @@
...
+
+class p21cmEMU():
```
steven-murray: no need for parentheses... use `class p21cmEMU:`

src/py21cmmc/emulator.py
```diff
@@ -0,0 +1,197 @@
...
+            log.warning('No emulator path was supplied. Trying to find the emulator folder in the current working directory.')
+            # Check all local directories
+            dirs = []
+            for i in os.listdir('.'):
+                if os.path.isdir(i):
+                    dirs.append(i)
+            mask = ['emu' in i.lower() for i in dirs]
+            if any(mask):
+                emu_path = np.array(dirs)[mask][0]
+                log.info('Emulator path found: %s' % emu_path)
```
steven-murray: Honestly, I think this kind of code is too brittle. Just require a path to be set, or have a single default path where the emulator should be (for instance, we could let the emulator file be downloaded to a certain location from GDrive the first time this is called)

src/py21cmmc/emulator.py
```diff
@@ -0,0 +1,197 @@
...
+    def predict(self, params):
+        r"""
+        Call the emulator, evaluate it at the given parameters, restore dimensions.
+    
+        Parameters
+        ----------
+        theta : p21.AstroParams or np.ndarray
+        An array with the nine astro_params input all $\in [0,1]$.
+        """
+        log.info('Beginning prediction...')
```
steven-murray: since prediction should be _fast_, I probably wouldn't include too many logging statements here (maybe none except for a potential `.debug()` statement)

src/py21cmmc/emulator.py
```diff
@@ -0,0 +1,197 @@
...
+        if isinstance(params, p21.AstroParams) or type(params) == dict:
```
steven-murray: I think this code will be slightly faster as follows:

```python
param_keys = ['F_STAR10','ALPHA_STAR','F_ESC10','ALPHA_ESC','M_TURN','t_STAR','L_X','NU_X_THRESH','X_RAY_SPEC_INDEX']
if isinstance(params, p21.AstroParams):
    theta = np.array([params.defining_dict[key] for key in param_keys])
elif isinstance(params, dict):
    theta = np.array([params[key] for key in param_keys])
```

steven-murray: Here we just evaluate the if-statement once, instead of nparams times.

src/py21cmmc/emulator.py
```diff
@@ -0,0 +1,197 @@
...
+        if isinstance(params, p21.AstroParams) or max(theta.ravel()) > 1 or min(theta.ravel()) < 0:
+            log.info('Theta is not normalized.')
+            normed = False # to indicate that input params was not normalised
+            for i in range(theta.shape[-1]):
+                if i == 0 or i == 2 or i == 4 or i == 6:
+                    theta[:,i] = np.log10(theta[:,i])
+                if i == 7:
+                    theta[:,i] /= 1000. #E0 in keV
+                theta[:,i] -= self.limits[i,0]
+                theta[:,i] /= (self.limits[i,1] - self.limits[i,0])
```
steven-murray: this doesn't need to be in a for-loop. Just do something like:

```python
theta[:, [0,2,4,6]] = np.log10(theta[:, [0,2,4,6]])
theta[:, 7] /= 1000
theta -= self.limits[:, 0]
theta /= (self.limits[:, 1] - self.limits[:, 0])
```

src/py21cmmc/core.py
```diff
@@ -12,7 +12,7 @@
 import warnings
 from os import path
 from scipy.interpolate import interp1d
-
+from .emulator import p21cmEMU
```
qyx268: just a thought on this. it's probably better to have a repo of 21cmEMU in parallel to 21cmmc and 21cmFAST. I imagine there will be a lot of different versions of 21cmEMU in the future...

src/py21cmmc/core.py
```diff
@@ -1151,3 +1151,120 @@ def get_cl(self, cosmo, l_max=-1):
...
+    -----
+    This core calls 21cmEMU and uses it to evaluate 21cmFAST summaries (power spectrum, global signal, neutral fraction, spin temperature)
+    given a set of astro_params.
+
+    Parameters
+    ----------
+    redshift : float or array_like
+         The redshift(s) at which to evaluate the summary statistics.
+    user_params : dict or :class:`~py21cmfast.UserParams`
+        Parameters affecting the overall dimensions of the cubes.
```
qyx268: not only, user_params also define things affecting hmf, IC, etc. I assume the emulator you provides here was trained using 1 set of user_params, flag_options and cosmo_params. It is good to keep these as input for 21cmEMU here for future expansion. But right now, if what I assumed is true, maybe put a warning/todo here so people will know these arguments do not impact the 21cmEMU outcomes

src/py21cmmc/core.py
```diff
@@ -1151,3 +1151,120 @@ def get_cl(self, cosmo, l_max=-1):
...
+        flag_options=None,
+        astro_params=None,
+        cosmo_params=None,
+        global_params=None,
+        ctx_variables=("brightness_temp", "brightness_temp_err",
+                       "spin_temp", "spin_temp_err", "xHI","redshifts", "ps_redshifts",
+                       "xHI_err", "delta", "Muv", "lfunc", "uv_lfs_redshifts",
+                       "delta_err", "k", "tau_e", "tau_e_err"),
+        write=None,
+        emu_path=None,
```
qyx268: come out with a default emu_path, e.g., a relative path that when people download the emulator's essentials, they will be put there (like `py21cmmc/21cmEMU`). Also, please add `k` and `emu_path` in the Notes above since they are new in 21cmmc.

src/py21cmmc/core.py
```diff
@@ -1151,3 +1151,120 @@ def get_cl(self, cosmo, l_max=-1):
...
+
+        self.astro_params = p21.AstroParams(astro_params)
+        self.cosmo_params = p21.CosmoParams(cosmo_params)
+
+        self.global_params = global_params or {}
+        
+        self.emu_path = emu_path
+        
+        self.io_options = {
+                "write": write} # If not None, emulator outputs will be written to this folder
```
qyx268: perhaps follow the default io_options like the other?
e.g.
                "store": {},  # (derived) quantities to store in the MCMC chain.
                "cache_dir": None,  # where full data sets will be written/read from.
                "cache_mcmc": False,  # whether to cache ionization data sets

src/py21cmmc/core.py
```diff
@@ -1151,3 +1151,120 @@ def get_cl(self, cosmo, l_max=-1):
...
+
+
+    def build_model_data(self, ctx):
+        """Compute all data defined by this core and add it to the context."""
+        # Update parameters
+        logger.debug(f"Updating parameters: {ctx.getParams()}")
+        astro_params, cosmo_params = self._update_params(ctx.getParams())
+        logger.debug(f"AstroParams: {astro_params}")
+        logger.debug(f"CosmoParams: {cosmo_params}")
+        ## TODO check that cosmo_params = default params (wtv Yuxiang used) and if they are not, return err msg
```
qyx268: This is what I was saying above. So there, because u haven't implemented supports for varying cosmologies, you can issue a warning if the given cosmology differs from your training and tell people cosmology is reset to what u were using

src/py21cmmc/core.py
```diff
@@ -1151,3 +1151,120 @@ def get_cl(self, cosmo, l_max=-1):
...
+
+        # Call 21cmEMU wrapper which returns a dict
+        all_summaries = p21cmEMU(emu_path=self.emu_path,
+                                 fname=self.io_options["write"]).predict(
+                                 params=astro_params,
```
qyx268: if you are not distinguishing astro_params, cosmo_params, etc in your params of p21cmEMU, maybe merge them here. in principle, if the future plan is to have these as input of the emulator, maybe rename params as astro_params, and leave space for the others?

src/py21cmmc/likelihood.py
```diff
@@ -994,18 +1000,36 @@ def computeLikelihood(self, model):
         lnl : float
             The log-likelihood for the given model.
         """
-        return (
```
qyx268, line 997: let's do it more consistently here. let's set an argument in `__init__` named e.g. flat_double_gaussian with a default value being False. Then change your comment `if self._is_emu` to `if self.flag_double_gaussian`

qyx268: as for emulator's tau_err, you can add them to 

                    self.tau_sigma_u * self.tau_sigma_l
                    + (self.tau_sigma_u - self.tau_sigma_l) * (model["tau"] - self.tau_mean)
                
too

src/py21cmmc/likelihood.py
```diff
@@ -1014,51 +1038,63 @@ def reduce_data(self, ctx):
         -------
         dict :
             Only key is "tau", the optical depth to reionization.
+            If the emulator is used, 'tau_err' is the emulator error on tau calculated using compute_tau.
         """
         # Extract relevant info from the context.
 
+        if self._is_emu:
+            redshifts = ctx.get('redshifts')
```
qyx268: no need for line 1046 and 1047

src/py21cmmc/likelihood.py
```diff
@@ -1014,51 +1038,63 @@ def reduce_data(self, ctx):
...
+            # Once again, performed using command line code.
+            # TODO: not sure if this works.
+            tau_value = lib.compute_tau(
+                user_params=self.core_primary.user_params,
+                cosmo_params=self.core_primary.cosmo_params,
+                redshifts=z_extrap,
+                global_xHI=xHI,
+            )
 
-        return {"tau": tau_value}
```
qyx268, line 1061: then here, you return `'tau_err'` even if not self._is_emu. but set it to zero

src/py21cmmc/likelihood.py
```diff
@@ -1168,30 +1218,47 @@ def reduce_data(self, ctx):
                 # 1.0818709330934035 = 1+0.25*YHe/(1-YHe) with YHe coming from BBN
                 # I think there is a way to fix YHe using user defined value
                 xHI = ctx.get("xHI")
+                if xHI.shape[0] == 1:
+                    xHI = xHI.flatten()
                 redshifts = ctx.get("zs")
+                if redshifts is None:
```
qyx268, line 1236: this is not a good strategy. did you name redshifts as redshifts instead of zs on purpose to do this? 

You can just put a if self_emu_modules at the beginning and put if self._use_coeval:BLABLA else: BLABLA inside a else. Then you don't need `if self.emu_modules` in setup either.

src/py21cmmc/likelihood.py
```diff
@@ -1168,30 +1218,47 @@ def reduce_data(self, ctx):
...
             else:
                 xHI = ctx.get("lightcone").global_xHI
                 redshifts = ctx.get("lightcone").node_redshifts
-
         redshifts, xHI = np.sort([redshifts, xHI])
-        return {"xHI": xHI, "redshifts": redshifts}
+        if err is None:
+            return {"xHI": xHI, "redshifts": redshifts}
+        else:
+            return {"xHI": xHI, "redshifts": redshifts, 'err': err}
```
qyx268: following what mentioned in tau_e, return err for both case but assign zero or zeros to it if not using emulator

src/py21cmmc/likelihood.py
```diff
@@ -1168,30 +1218,47 @@ def reduce_data(self, ctx):
...
         """Compute the likelihood."""
         lnprob = 0
-
+                    
+                    
         if self._require_spline:
             model_spline = InterpolatedUnivariateSpline(
                 model["redshifts"], model["xHI"], k=1
             )
+            if 'err' in model.keys():
```
qyx268: Then here, it can be something  like if np.sum( model["err"]) < 1e-3

src/py21cmmc/likelihood.py
```diff
@@ -1289,15 +1356,22 @@ def computeLikelihood(self, model):
...
+    required_cores = ((core.CoreLightConeModule,core.Core21cmEMU),)
 
     def reduce_data(self, ctx):
         """Reduce data to model."""
-        return {
-            "frequencies": 1420.4
-            / (np.array(ctx.get("lightcone").node_redshifts) + 1.0),
-            "global_signal": ctx.get("lightcone").global_brightness_temp,
-        }
+        if isinstance(self._cores[0], core.Core21cmEMU):
```
qyx268: can u check if self._cores[0] is indeed Core21cmEMU? Doesn't self._cores contain many cores (e.g. LFs, etc). You might need to do `emu_modules` as in LikelihoodNeutralFraction. 

If you are sure this is fine, then you don't need `emu_modules` in LikelihoodNeutralFraction

src/py21cmmc/likelihood.py
```diff
@@ -1425,43 +1505,63 @@ def paired_core(self):
         """The luminosity function core that is paired with this likelihood."""
         paired = []
         for c in self._cores:
-            if isinstance(c, core.CoreLuminosityFunction) and c.name == self.name:
+            if (isinstance(c, core.CoreLuminosityFunction) and c.name == self.name) or (isinstance(c, core.Core21cmEMU) and c.name == self.name):
```
qyx268: does this mean that you are specifying names in `Core21cmEMU` to match `names` used in this likelihood, just like what we are doing with CoreLuminosityFunction? This means you will call Core21cmEMU multiple times if you have multiple `LikelihoodLuminosityFunction`, which will waste a lot, no?

src/py21cmmc/likelihood.py
```diff
@@ -1425,43 +1505,63 @@ def paired_core(self):
...
     def redshifts(self):
         """Redshifts at which luminosity function is defined."""
-        return self.paired_core.redshift
+        if isinstance(self.paired_core, core.Core21cmEMU):
+            if self.z is not None:
+                if hasattr(self.z, "__len__"):
+                    return self.z
+                else:
+                    return np.array([self.z])
+            else:
```
qyx268, line 1536: under this else, what would happen to line 1442?

src/py21cmmc/likelihood.py
```diff
@@ -1425,43 +1505,63 @@ def paired_core(self):
...
+                # list of length one of datafile to be written.
+                return [lfunc]
+            else:
+                return lfunc
 
     def computeLikelihood(self, model):
         """Compute the likelihood."""
         lnl = 0
         for i, z in enumerate(self.redshifts):
+            # mask to remove possible NaNs returned by p21.wrapper.compute_luminosity_function()
```
qyx268: has this ever happened?

src/py21cmmc/likelihood.py
```diff
@@ -1782,3 +1882,136 @@ def computeLikelihood(self, model):
             )
             return -np.inf
         return lnl
+
+class Likelihood1DPowerLightconeUpper(Likelihood1DPowerLightcone):
+    r"""
+    Likelihood based on Chi^2 comparison to luminosity function data.
```
qyx268: 21cmPS

src/py21cmmc/likelihood.py
```diff
@@ -1782,3 +1882,136 @@ def computeLikelihood(self, model):
...
+            self.k_len = len(k)
+        if datafile is not None:
+            self.datafile = [datafile]
+            self.data = dict(np.load(datafile))
+            if k is None:
+                self.k = [self.data['kwfband10'], self.data['kwfband8']]
+                self.k_len = max(len(i) for i in self.k)
+            
+
+    @classmethod
```
qyx268, line 1939: i'm not familiar with this. But why not using `_read_data` in `LikelihoodBaseFile`?

src/py21cmmc/likelihood.py
```diff
@@ -1782,3 +1882,136 @@ def computeLikelihood(self, model):
...
+                else:
+                    error_val = np.sqrt(PS_limit_vars + (0.2 * ModelPS_val_afterWF)**2)
+                print('old', np.sqrt(PS_limit_vars + (0.2 * ModelPS_val_afterWF)**2))
+                likelihood = 0.5 + 0.5 * erf( ( PS_limit_vals - ModelPS_val_afterWF ) / (np.sqrt(2) * error_val) ) # another way to write likelihood for 1-side Gaussian
+                likelihood[likelihood <= 0.0] = 1e-50
+                lnl += np.nansum(np.log(likelihood))
+                
+        return lnl
+    
+    @cached_property
```
qyx268, line 2060: why r u pairing? same as before with luminosity function. I dont think u need to pair under the current structure of how u use the emualtor

src/py21cmmc/likelihood.py
```diff
@@ -1782,3 +1882,136 @@ def computeLikelihood(self, model):
...
+        * ``kwfbandx``: the k values [/Mpc] of the window function for band x
+    """
+
+    def __init__(self, datafile="", data = None, name="", 
+                 redshifts = np.array([10.37213048, 7.92876298]),
+                 k=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.name = name
+        self.redshifts = redshifts
+        if k is not None:
```
qyx268: You current setup doesn't actually allow u to specify k by yourself, because the window function is fixed at give ks...

src/py21cmmc/likelihood.py
```diff
@@ -1782,3 +1882,136 @@ def computeLikelihood(self, model):
...
+    datafile : str, optional
+        Input data should be in a `.npz` file, and contain the arrays:
+        * ``z_bands``: the redshift of the observation bands
+        * ``bandx``: the power spectrum upper limits at redshift x
+        * ``wfbandx``: the window function for band x
+        * ``kwfbandx``: the k values [/Mpc] of the window function for band x
+    """
+
+    def __init__(self, datafile="", data = None, name="", 
+                 redshifts = np.array([10.37213048, 7.92876298]),
```
qyx268: same as redshift, you cant really change the redshift to other values, in particular when they are far away from 8 and 10

src/py21cmmc/likelihood.py
```diff
@@ -1782,3 +1882,136 @@ def computeLikelihood(self, model):
...
+                PS_limit_ks = self.data[0]['band' + str(round(band))][field,:,0]
+                PS_limit_ks = PS_limit_ks[~np.isnan(PS_limit_ks)]
+                Nkbins = len(PS_limit_ks)
+                PS_limit_vals = self.data[0]['band' + str(round(band))][field,:Nkbins,1] 
+                PS_limit_vars = self.data[0]['band' + str(round(band))][field,:Nkbins,2]  
+
+                kwf_limit_vals = self.data[0]['kwfband' + str(round(band))]
+                Nkwfbins = len(kwf_limit_vals)
+                PS_limit_wfcs = self.data[0]['wfband' + str(round(band))][field,:Nkbins,:]
+                mask = ~np.isnan(PS_limit_wfcs)
```
qyx268: does the mask really do anything here? If so, you won't be able to reshape below

src/py21cmmc/likelihood.py
```diff
@@ -1782,3 +1882,136 @@ def computeLikelihood(self, model):
...
+                PS_limit_wfcs = PS_limit_wfcs[mask].reshape([Nkbins, Nkwfbins])
+                
+                model_ks = model[0]["k"]
+                model_zs = self.redshifts
+                zbin = np.argmin(abs(band - model_zs))
+                ModelPS_val = model[0]["delta"][zbin,:Nkwfbins]
+
+                if np.sum(np.isnan(PS_limit_wfcs)) != 0:
+                    mask1 = ~np.isnan(PS_limit_wfcs)
```
qyx268: mask1 is not used

src/py21cmmc/likelihood.py
```diff
@@ -1782,3 +1882,136 @@ def computeLikelihood(self, model):
...
+                    #error_val_upper = np.sqrt(PS_limit_vars + (0.2*ModelPS_val_afterWF)**2 + (ModelPS_val_1sigma_upper_afterWF-ModelPS_val_afterWF)**2 ) 
+                    #error_val_lower = np.sqrt(PS_limit_vars + (0.2*ModelPS_val_afterWF)**2 + (ModelPS_val_1sigma_lower_afterWF-ModelPS_val_afterWF)**2 )
+                    #mask = PS_limit_vals > ModelPS_val_afterWF
+                    #error_val = error_val_lower
+                    #error_val[mask] = error_val_upper[mask]
+                    mean_err = np.mean([ModelPS_val_1sigma_upper_afterWF-ModelPS_val_afterWF, ModelPS_val_afterWF-ModelPS_val_1sigma_lower_afterWF], axis = 0)
+                    error_val = np.sqrt(PS_limit_vars + (0.2*ModelPS_val_afterWF)**2 + (mean_err)**2 )
+                else:
+                    error_val = np.sqrt(PS_limit_vars + (0.2 * ModelPS_val_afterWF)**2)
+                print('old', np.sqrt(PS_limit_vars + (0.2 * ModelPS_val_afterWF)**2))
```
qyx268: remove or put in debug

src/test_21cmmc.py
```diff
@@ -0,0 +1,30 @@
+from py21cmmc import mcmc, core, analyse, likelihood
```
qyx268: These should be moved into 21CMMC/tests

src/py21cmmc/emulator.py
```diff
@@ -0,0 +1,196 @@
...
+        if len(theta.shape) == 1:
+            theta = theta.reshape([1,-1])
+        normed = True
+        # Check that theta is normalized, if not, normalise it.
+        if isinstance(params, p21.AstroParams) or max(theta.ravel()) > 1 or min(theta.ravel()) < 0:
+            log.info('Theta is not normalized.')
+            normed = False # to indicate that input params was not normalised
+            for i in range(theta.shape[-1]):
+                if i == 0 or i == 2 or i == 4 or i == 6:
+                    theta[:,i] = np.log10(theta[:,i])
```
qyx268: double check whether these thetas are already in log space or not

src/py21cmmc/emulator.py
```diff
@@ -0,0 +1,196 @@
...
+                    theta[:,i] /= 1000. #E0 in keV
+                theta[:,i] -= self.limits[i,0]
+                theta[:,i] /= (self.limits[i,1] - self.limits[i,0])
+        log.info('Normed params: %s' % theta)
+        log.info('Start emulator prediction...')
+        emu_pred = self.model.predict(theta, verbose = False)
+        log.info('End emulator prediction.')
+
+        Tb_pred_normed = emu_pred[:,:84]
+        xHI_pred = emu_pred[:,84:84*2]
```
qyx268: There are many hard-coded numbers, which I'm not checking. please make sure they are right.

qyx268: better to have a comment next to each line, saying what emu_pred[:BLA to BLA] is

src/py21cmmc/emulator.py
```diff
@@ -0,0 +1,196 @@
...
+        PS_pred_normed = emu_pred[:,84*3+1:].reshape((theta.shape[0], 60,12))
+
+        # Set the xHI < z(Ts undefined) to 0
+        xHI_pred_fix = np.zeros(xHI_pred.shape)
+
+        tau = np.zeros(theta.shape[0])
+        uvlfs = np.zeros((theta.shape[0], 3, len(self.uv_lf_zs),100))
+        for i in range(theta.shape[0]):
+            zbin = np.argmin(abs(self.zs - Ts_undefined_pred[i]))
+            if xHI_pred[i,zbin] < 1e-1:
```
qyx268: I dont understand what this part is about...

src/py21cmmc/emulator.py
```diff
@@ -0,0 +1,196 @@
...
+            # Use py21cmFAST to analytically calculate UV LF and $\tau_e$
+            log.info('Begin tau computation...')
+            tau[i] = p21.wrapper.compute_tau(redshifts = self.zs, global_xHI = xHI_pred_fix[i,:])
+            log.info('Tau computation completed successfully.')
+
+            # Build astro_params array from theta \in [0,1]
+            if (isinstance(params, p21.AstroParams) or type(params) == dict) and not normed:
+                uvlfs[i,...] = np.array(p21.wrapper.compute_luminosity_function(redshifts = self.uv_lf_zs, astro_params=params))
+            else:
+                # Restore dimensions i.e. undo the limits
```
qyx268: why not calculate luminosity function before normlizing?

src/py21cmmc/emulator.py
```diff
@@ -0,0 +1,196 @@
...
+                log.info('Restore dimensions to theta to calculate UV LF.')
+                theta_wdims = np.zeros(theta.shape)
+                for j in range(theta.shape[-1]):
+                    theta_wdims[:,j] = self.limits[j, 0] + theta[:, j] * (self.limits[j, 1] - self.limits[j, 0])
+                    if j == 7:
+                        theta_wdims[:,j] = 1000. * theta_wdims[:,j]
```
qyx268: also, I thought u are emulating tau and UVLF. But it turns out you are not. Then why go all the trouble modifying them in likelihood.py? Just do the calculation there

qyx268 (commented):
several major comments:
1. in principle, we shouldn't change likelihoods too much. We should alter cores to allow them to get e.g. xHI(z) or LFs or 21cmPS based on whether or not we are using emulator. e.g. in `CoreLuminosityFunction`, if Core21cmEMU if also specified, then in `run` you call the emulator instead of calling `compute_luminosity_function`. This will require major changes to your branch. You can consider doing it in the future. 
2. Basically, many likelihoods in likelihood.py can be updated to use your emulator, e.g. Likelihood1DPowerLightcone, LikelihoodEDGES, etc. In particular, Likelihood1DPowerLightconeUpper should support scenarios when emulator is not asked and can be merged into Likelihood1DPowerLightcone. This can be left to future work when they are needed, which will be soon I believe.

src/py21cmmc/core.py
DanielaBreitman: Just to be sure can you tell me what that cosmology should be? It's just Planck18, no?

qyx268: SIGMA_8    0.82
hubble    0.6774
Omega_M    0.3075
Omega_L    0.692500
Omega_b    0.0486
ns    0.97

The point is that cosmo_params can be part of your network feature.

src/py21cmmc/emulator.py
DanielaBreitman: Because Andrei wants them returned by the emulator (yes, even though they are not emulated). So since the calculation is already done inside, I don't want to do it twice.

src/py21cmmc/emulator.py
DanielaBreitman: Quadruple checked already. The input to AstroParams is in logspace, but not when you read them out of the dict.

src/py21cmmc/emulator.py
DanielaBreitman: The emulator predicts the redshift at which Ts becomes undefined. This is also approx the same redshift at which xHI becomes 0, so this part says if the xHI value at this bin where Ts is not defined anymore is less than 0.1, then set every xHI at lower z to 0.

src/py21cmmc/emulator.py
DanielaBreitman: I am. This is what line 137 checks for.

src/py21cmmc/core.py
DanielaBreitman: There will be a repo.

src/py21cmmc/core.py
DanielaBreitman: The current default path is '.' i.e. in the same directory as the code is run.

src/py21cmmc/core.py
DanielaBreitman: Yes I am adding it. Do you also have the flag_options and user_params from your runs? I can add that too.

src/py21cmmc/core.py
DanielaBreitman: Fixed

src/py21cmmc/core.py
DanielaBreitman: I wanted to distinguish the 21cmFAST folders from the emu folder, but I guess nobody will use the emulator and 21cmFAST simultaneously so this is probably fine. Fixed.

src/py21cmmc/core.py
DanielaBreitman: For now I put flag options all False

src/py21cmmc/core.py
DanielaBreitman: Fixed

src/py21cmmc/likelihood.py
DanielaBreitman: Fixed

src/py21cmmc/likelihood.py
DanielaBreitman: Fixed

src/py21cmmc/likelihood.py
DanielaBreitman: Fixed

src/py21cmmc/likelihood.py
DanielaBreitman: I did it the second way at the end i.e. add emulator error to the upper and lower errors of tau.

src/py21cmmc/likelihood.py
```diff
@@ -6,13 +6,15 @@
...
+
+from scipy.special import erf
+from scipy.interpolate import InterpolatedUnivariateSpline, interp1d, interp2d
 
 from . import core
 
 loaded_cliks = {}
 logger = logging.getLogger("21cmFAST")
-
+logger.setLevel(logging.DEBUG)
```
steven-murray, line 17: don't set the level in a library module!

src/py21cmmc/likelihood.py
```diff
@@ -118,14 +120,16 @@ class LikelihoodBaseFile(LikelihoodBase):
...
+        Default is False
     use_data : bool, optional
         Sometimes you may want to construct the likelihood without any data at all.
         Set this to False if this is the case.
     """
 
     _ignore_attributes = LikelihoodBase._ignore_attributes + ("simulate",)
 
-    def __init__(self, datafile=None, noisefile=None, simulate=False, use_data=True):
+    def __init__(self, datafile=None, noisefile=None, simulate=False, emulate=False, use_data=True):
```
steven-murray: It does not seem useful to have the `emulate` parameter -- it will emulate if you have `simulate=True` and the Emulator in your Core.

src/py21cmmc/likelihood.py
steven-murray: I think I agree

steven-murray (commented):
Thanks @DanielaBreitman this is a lot of awesome work! I have a few comments, and hopefully after they are addressed,  I can have another finer-grained look!

src/py21cmmc/core.py
qyx268: The two versions of 21cmFAST are not identical. But I think you can safely use:
```
        "USE_HALO_FIELD": False,
        "USE_MINI_HALOS": False,
        "USE_MASS_DEPENDENT_ZETA": True,
        "SUBCELL_RSD": True,
        "INHOMO_RECO": True,
        "USE_TS_FLUCT": True,
        "M_MIN_in_Mass": False,
        "PHOTON_CONS": True,
        "FIX_VCB_AVG": False,
        "EVOLVING_R_BUBBLE_MAX": False,
```
and 
```
        "BOX_LEN": 250,
        "DIM": 512,
        "HII_DIM": 128,
        "USE_FFTW_WISDOM": True,
        "HMF": 1,
        "USE_RELATIVE_VELOCITIES": False,
        "POWER_SPECTRUM": 0,
        "N_THREADS": 1,
        "PERTURB_ON_HIGH_RES": False,
        "NO_RNG": False,
        "USE_INTERPOLATION_TABLES": True,
        "FAST_FCOLL_TABLES": False,
        "USE_2LPT": True,
        "MINIMIZE_MEMORY": False,

```

DanielaBreitman:
Emulator tests do not run on this branch (serialization error when loading the emulator tf model). Switching to another branch (where everything runs smoothly).

PR #74, opened 2023-07-18
Emulator integration + HERA H1C PS Upper limit likelihood

DanielaBreitman:
Adding 

- Emulator support for LF, neutral fraction, and Planck (tau) likelihoods.
- HERA H1C PS upper bound likelihood (emulator only, no 21cmFAST support yet)


src/py21cmmc/core.py
```diff
@@ -1151,3 +1151,138 @@ def get_cl(self, cosmo, l_max=-1):
...
+        astro_params = self._update_params(ctx.getParams())
+        logger.debug(f"AstroParams: {astro_params}")
+        # Take only needed AstroParams
+        input_dict = {k: getattr(astro_params, k)
+                for k in self.astro_param_keys}
+
+        # Call 21cmEMU wrapper which returns a dict
+        theta, outputs, errors = self.emulator.predict(astro_params=input_dict)
+        if self.io_options['cache_dir'] is not None:
+            outputs.write(fname = self.io_options['cache_dir'],#Does this include the fname as well?
```
steven-murray: Typically the `cache_dir` would not include the filename, only the directory. I think the idea is that the fname might be specified dynamically.

src/py21cmmc/likelihood.py
```diff
@@ -4,15 +4,17 @@
...
-from scipy.interpolate import InterpolatedUnivariateSpline
+from scipy.interpolate import InterpolatedUnivariateSpline, interp1d, interp2d
+from scipy.special import erf
 
 from . import core
 
 loaded_cliks = {}
 logger = logging.getLogger("21cmFAST")
-
+logger.setLevel(logging.DEBUG)
```
steven-murray: Don't set logging levels in library code. Only in scripts / CLI's etc.

src/py21cmmc/likelihood.py
```diff
@@ -1357,12 +1441,20 @@ def __init__(self, *args, name="", mag_brightest=-20.0, **kwargs):
...
         self.mag_brightest = mag_brightest
 
     def setup(self):
         """Setup instance."""
+        if isinstance(self.paired_core, core.Core21cmEMU):
+            if self.z is None:
+                raise ValueError(
+                    "Must specify which z bin out of [6, 7, 8, 10] the likelihood is comparing to the data."
+                )
+            self.i = np.argmin(abs(np.array([6, 7, 8, 10]) - int(self.z)))
```
steven-murray: I think this is a poor choice of attribute name. Maybe `z_index`?

src/py21cmmc/likelihood.py
```diff
@@ -1782,3 +1891,181 @@ def computeLikelihood(self, model):
...
+    ):
+        super().__init__(*args, **kwargs)
+        self.name = name
+        self.redshifts = redshifts
+
+        if not hasattr(self.redshifts, "__len__"):
+            self.redshifts = np.array([self.redshifts])
+        if len(self.redshifts) > 2:
+            raise ValueError(
+                "The HERA H1C data only contains two redshift bands. You supplied %s redshift bands."
```
steven-murray: It's not clear to me why you're restricting this likelihood to be compatible with exactly HERA H1C -- why not let it read any data and just use the redshifts defined in that data?

tests/test_emulator.py
```diff
@@ -0,0 +1,8 @@
+def test_emulator_install():
+    from py21cmemu import Emulator
+    emu = Emulator()
+
+def test_emulator_runs():
+    from py21cmemu import Emulator
```
steven-murray: I don't think you need these tests -- these tests should be just in the py21cmemu package (which they are).

steven-murray (commented):
Thanks @DanielaBreitman, this is looking good. Other than the minor comments I made on specific lines, I have two more general comments:

1. The upper-limit likelihood that you wrote is very specific to HERA H1C, but we'll end up using the same likelihood for H4C and maybe H6C too, so why make it so specific to those redshifts etc? Why not let it just read a HERA datafile with any number of redshifts in it, and then use those redshifts?
2. `pspec_likelihood` defines the upper-limit likelihood, and also applies window functions etc properly to HERA-like data. Why not just use it in your likelihood?

Also, you'll need to get the pre-commit failures fixed. I think it's just a matter of changing the .flake8 file to have comments on their own lines instead of after their ignores. Eg. instead of

```
F402  # this does something
```

have 

```
# this does something
F402
```

tests/test_emulator.py
DanielaBreitman: Just to make sure that 21cmEMU is installed properly e.g. they were failing before when I had the weird tensorflow bug. The tests of py21cmemu are not ran upon install, so I thought this would be useful to make sure it's installed properly. But I can remove them if you think it's not necessary.

DanielaBreitman:
Hi @steven-murray, concerning the pre-commit failure: how did you come to this conclusion? The error shows some issue with the poetry build (see [here)](https://results.pre-commit.ci/run/github/191839335/1689684106.4GqD2J77TsqclSjoR02qQg). I did move the comments in .flake8 to a separate line like you suggested though and that does not fix it. 

After some extra searching, this build issue is a problem with the `isort` version (see [here](https://stackoverflow.com/questions/75281731/pre-commit-fails-to-install-isort-5-10-1-with-error-runtimeerror-the-poetry-co))

DanielaBreitman:
Fix for the pre-commit build issue was just to run `pre-commit autoupdate` (which updates the `isort` version to 5.12.0 where this issue does not occur anymore)

DanielaBreitman:
@steven-murray  One issue I realised. The DMI has the theory_model argument as part of its input that takes in a theory model function, not the evaluated theory model. In 21CMMC, computeLikelihood contains an evaluated model, not a model function. So I would have to make changes to the DMI class if you want to integrate it into this code. With this in mind, it might be better to leave the likelihood calculation as is?

tests/test_emulator.py
steven-murray: Yeah, I think having the other test -- of the likelihood itself -- is enough to show that the package is installed.

steven-murray:
@DanielaBreitman yes the DMI uses a functional form of the theory model -- this is to allow it to evaluate at the kperp/kpar of the data, rather than that of the model (if such a thing exists). You just need to do an interpolation to make this work.

steven-murray (approved):
This looks good, thanks @DanielaBreitman. Just a note that in the future we'll swap out the likelihood guts to use `pspec_likelihood`.

PR #75, opened 2023-08-10
Implement emulator support and vecorization for the likelihood Likelihood1DPowerLightCone

DanielaBreitman:


steven-murray (approved):
Thanks @DanielaBreitman I think this looks good!

PR #76, opened 2023-08-28
Ultranest sampler

dprelogo:
adding Ultranest sampler option


.npz
steven-murray, line 1: I think this file was added by accident?

src/py21cmmc/core.py
```diff
@@ -607,38 +613,89 @@ def run(self, astro_params, cosmo_params, ctx):
...
-            return p21.compute_luminosity_function(
-                mturnovers=mturnovers,
-                mturnovers_mini=mturnovers_mini,
-                redshifts=self.redshift,
-                astro_params=astro_params,
-                flag_options=self.flag_options,
-                cosmo_params=cosmo_params,
-                user_params=self.user_params,
-                nbins=self.n_muv_bins,
+            if type(astro_params) == np.ndarray:
```
steven-murray: use `isinstance(astrop_params, np.ndarray)` instead

src/py21cmmc/core.py
```diff
@@ -607,38 +613,89 @@ def run(self, astro_params, cosmo_params, ctx):
...
+                N = 1
+            Muv = []
+            Mhalo = []
+            lfunc = []
+            for i in range(N):
+                muv, mhalo, lf = p21.compute_luminosity_function(
+                    mturnovers=mturnovers,
+                    mturnovers_mini=mturnovers_mini,
+                    redshifts=self.redshift,
+                    astro_params=astro_params,
```
steven-murray: should `astro_params` here by indexed by `i`?

src/py21cmmc/core.py
steven-murray: I *think* this would be cleaner if you set the `mturnovers` and `mturnovers_mini` to None if NOT doing mini halos, and then out-dented the `compute_luminosity_function` call.

src/py21cmmc/core.py
```diff
@@ -607,38 +613,89 @@ def run(self, astro_params, cosmo_params, ctx):
...
             )
         else:
-            return p21.compute_luminosity_function(
-                redshifts=self.redshift,
-                astro_params=astro_params,
-                flag_options=self.flag_options,
-                cosmo_params=cosmo_params,
-                user_params=self.user_params,
-                nbins=self.n_muv_bins,
+            if type(astro_params) == np.ndarray:
```
steven-murray: Maybe we should note somewhere in the documentation for this class and/or method that you can pass `astro_params` as an array now.

src/py21cmmc/likelihood.py
```diff
@@ -1429,33 +1429,53 @@ def reduce_data(self, ctx):
...
                 redshifts = ctx.get("lightcone").node_redshifts
-        redshifts, xHI = np.sort([redshifts, xHI])
+        if len(xHI.shape) == 1:
+            redshifts, xHI = np.sort([redshifts, xHI])
         return {"xHI": xHI, "redshifts": redshifts, "err": err}
 
     def computeLikelihood(self, model):
         """Compute the likelihood."""
-        lnprob = 0
+        if len(model["xHI"].shape) == 1:
```
steven-murray: instead use `model['xHI'].ndim`

steven-murray: In fact, might be better to use

```python
xHI = np.atleast2d(model['xHI'])
n = xHI.shape[0]
```

then you don't need the extra check for `n==1` below.

src/py21cmmc/likelihood.py
```diff
@@ -1429,33 +1429,53 @@ def reduce_data(self, ctx):
...
+                    if n == 1:
+                        lnprob += self.lnprob(
+                            model["xHI"][model["redshifts"].index(z)], data, sigma_t
+                        )
+                    else:
+                        lnprob[i] += self.lnprob(
+                            model["xHI"][model["redshifts"].index(z)], data, sigma_t
+                        )
+                else:
+                    if n == 1:
```
steven-murray: Again, I think just have `lnprob` always be an array, even if it's just size 1.

src/py21cmmc/likelihood.py
```diff
@@ -1631,18 +1690,18 @@ def __init__(self, *args, name="", mag_brightest=-20.0, z_idx=None, **kwargs):
...
     def setup(self):
         """Setup instance."""
         if isinstance(self.paired_core, core.Core21cmEMU):
-            if self.z_idx is None:
+            if self.z is None:
                 raise ValueError(
                     "Must specify which z bin out of [6, 7, 8, 10] the likelihood is comparing to the data."
                 )
-            self.i = np.argmin(abs(np.array([6, 7, 8, 10]) - int(self.z_idx)))
+            self.i = np.argmin(abs(np.array([6, 7, 8, 10]) - int(self.z)))
```
steven-murray: Please don't use `i` as an attribute name. Use `zidx` or something instead.

src/py21cmmc/likelihood.py
```diff
@@ -1746,19 +1817,50 @@ def reduce_data(self, ctx):
...
-        lnl = 0
-        for i, z in enumerate(self.redshifts):
-            model_spline = InterpolatedUnivariateSpline(
-                model["Muv"][i][::-1], model["lfunc"][i][::-1]
-            )
+        N = model["lfunc"].shape[0]
+        if N > 1:
+            lnl = np.zeros(N)
+        else:
+            lnl = 0
```
steven-murray: Again, just always make `lnl` an array.

steven-murray (commented):
Thanks @DanielaBreitman, this is looking good. I just made a few comments to clean things up a little.

DanielaBreitman (approved):

PR #79, opened 2023-11-20
fix:get correct z key

DanielaBreitman:


src/py21cmmc/likelihood.py
```diff
@@ -526,7 +526,7 @@ def computeLikelihood(self, model):
             for i in range(N):
                 all_band_keys = []
                 for key in list(hera_data.keys()):
-                    if "band" in key and "wf" not in key and "k" not in key:
+                    if "band" in key and "wf" not in key and "k" not in key and  and str(int(np.round(self.redshift))) in key:
```
steven-murray: typo (two "and"s here)

DanielaBreitman: I know gimme a minute...

steven-murray (approved):

PR #81, opened 2024-04-10
Fix non-vectorized case w 21cmEMU

DanielaBreitman:
Fix error when setting `vectorized = False` in the MCMC options while using `21cmEMU`.


src/py21cmmc/core.py
```diff
@@ -1329,9 +1326,13 @@ def build_model_data(self, ctx):
             for t in zip(*values):
                 ap.append(dict(zip(keys, t)))
             astro_params = np.array(ap, dtype=object)
+        if all(isinstance(v, (float, int)) for v in values) and len(values) > 0:
```
steven-murray: this should probably be an `elif`

steven-murray (approved):
Great! I just made one comment, but then it's fine

PR #84, opened 2024-06-25
Fix mpl and multithreading

disco-research2:
1) Use modern matplotlib approach to register colormaps
2) Fix multiprocessing (might be an issue with local hardware, please test before merging this!)

### DIFF
--- a/src/py21cmmc/mcmc.py
+++ b/src/py21cmmc/mcmc.py
@@ -577,6 +577,7 @@ def run_mcmc(
         return sampler, result
 
     else:
+        # didn't spawn Pools in my case, setting mp_context (either spawn or fork) fixed it
         pool = mcmc_options.pop(
             "pool",
             ProcessPoolExecutor(max_workers=mcmc_options.get("threadCount", 1), mp_context='spawn'),
--- a/src/py21cmmc/mcmc.py
+++ b/src/py21cmmc/mcmc.py
@@ -579,7 +579,7 @@ def run_mcmc(
     else:
         pool = mcmc_options.pop(
             "pool",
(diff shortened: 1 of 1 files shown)


steven-murray:
Thanks @Kojobu for the PR. I think the modern MPL part is a 21cmFAST thing, not 21cmMC -- there's actually an open PR there that fixes it. The `mp_context` thing I'll have to look into and make sure it doesn't mess things up for other hardware. Thanks again!

disco-research2:
The multiprocessing seems to be a local problem, you may skip this. The small fix for matplotlib may be worth a look though.

steven-murray:
@Kojobu OK thanks for noting that. The fix for matplotlib is not actually in this PR (see the changed files tab above). The problem is in 21cmFAST.

disco-research2:
@steven-murray You are right, my bad. So, I'm on Linux, Python version 3.11 and I consistently get memory leaks when running with INHOMO_RECO and TS_FLUCT. After too many trial and errors I found it's some issue with Python's concurrent.futures.ProcessPoolExecutor. 
I fixed it by changing the Pool in mcmc.py to
```python
set_start_method('spawn')
pool = mcmc_options.pop(
    "pool",
    ProcessPoolExecutor(
        max_workers=mcmc_options.get("threadCount", 1),
        max_tasks_per_child=1,
    ),
)
```
What this does is, it explicitly uses the spawn method (fork, which is standard in Linux, will be removed in Python3.14 according to the docs anyway) which creates a new process every run. It's kind of a brute force approach admittedly but performance-wise it shouldn't be much slower and it works.
I'm not sure if it's worth merging with main because it messes up with the logs a bit but it may be a workaround for future scientists running into the same issue. 
I think that's all from my side, thanks for the help! It's much appreciated. :)

